PR #15, opened 2026-08-26
Make the served embodiment an explicit launch flag

wong-1994:
## What

Makes the served embodiment an explicit launch flag, so an openpi-compatible
client and this server agree on the wire contract instead of both assuming
LIBERO.

Today `scripts/pi05_openpi_websocket_server.py` hard-codes LIBERO's keys
(`observation/image`, `observation/wrist_image`, `observation/state`). A
checkpoint fine-tuned for another robot is served silently against the wrong
dialect: the keys miss, the state routing differs, the action width differs, and
the failure shows up as degraded actions rather than an error. openpi solves this
by fixing the embodiment at startup via `--policy.config <TrainConfig>`; this PR
adds the equivalent `--robot`.

## Changes

**`--robot <preset>`** — a preset carries the camera wire keys (order-significant:
key *i* fills model view slot *i*), the state key, whether state is discretized
into the prompt, and the deployable action width. Two presets ship:

| preset | keys | state | action_dim |
|---|---|---|---|
| `franka_libero` (default, alias `libero`) | `observation/image`, `observation/wrist_image` | dropped | 7 |
| `unitree_g1` | `images/cam_high`, `images/cam_left_wrist`, `images/cam_right_wrist` | discretized | 32 |

Presets are named `<arm>_<key convention>` on purpose: LIBERO and DROID are both
Franka Panda with different keys and action spaces, so `franka` alone would not
fix the contract, and `libero` names a benchmark rather than a robot.

**Nested observation keys** — `apxinf.processors.transforms` gains
`lookup_key`/`has_key`/`set_key`, resolving flat-with-slash first
(`"observation/image"`, LIBERO/DROID) and then walking a nested path
(`obs["images"]["cam_high"]`, ALOHA/Unitree G1). Both wire shapes exist in the
wild; this handles them without the caller knowing which it is.

**`--num-views N`** — serves a checkpoint with fewer cameras than it declares.
This is numerically equivalent to what openpi does by zero-padding the absent
views and masking them out, but it skips their ~256 patch tokens per view
instead of computing and discarding them (measured 1.38x on a 2-view LIBERO
checkpoint served with one camera). The equivalence holds because masked tokens
consume no RoPE position (`positions = cumsum(input_mask) - 1`), are excluded
from attention (`valid_mask = mask[:, None, :] * mask[:, :, None]`), and
`embed_prefix` runs the same `PaliGemma.img` encoder per view with no per-slot
learned embedding. `num_views` only sizes the prefix; nothing weight-shaped
depends on it, so the load-time override is sound.

The flag is **required to be explicit** — a short `--image-keys` list on its own
is an error. A forgotten camera key should fail at startup, not quietly cost
accuracy.

**Loud failures** — a missing camera key raises with the served contract in the
message; `--num-views` above the checkpoint's own count is rejected at load;
`--num-views` disagreeing with the number of image keys is rejected at parse;
and `--random-weights` now warns that its synthetic tokenizer cannot honour a
preset's `discrete_state=True`, rather than publishing a contract it is not
serving.

**`apxinf_py.pyi`** — the stub was missing `action_horizon` and `sampling_seed`;
restored to match the Rust signature, with `num_views` added.

## Compatibility

`--num-views` is a server-side launch flag, not part of the wire protocol —
an unmodified openpi client is unaffected. `--robot` defaults to
`franka_libero`, which is byte-for-byte the previous behaviour, so existing
launches are unchanged.

## Verification

On Jetson AGX Orin (sm87), against this branch rebased onto `main`:

- `cargo check` and `cargo build --release -p apxinf-py --features cuda,extension-module` — clean
- 20/20 unit tests in `tests/test_robot_presets.py` (key resolution, preset validation, G1 serving, image slot order)
- Real checkpoint `pi05_libero_base` under `--robot franka_libero`: actions `(50, 7)`, finite, 2962 ms
- Same checkpoint with `--num-views 1`: actions `(50, 7)`, finite, 2140 ms (1.38x)
- `--num-views 4` on a 2-view checkpoint: rejected at load with the view count in the message
- Missing wrist camera: `KeyError` naming the served `image_keys`
- `--robot unitree_g1` over real websocket transport with an **unmodified** `openpi_client`: nested-key round trip returns `(50, 32)`; a LIBERO-shaped observation is rejected rather than served

### DIFF
--- a/README.md
+++ b/README.md
@@ -453,6 +453,15 @@ subprocess), so activate the environment built in
 source .venv/bin/activate
 ```
 
+`--robot` selects the **embodiment** — the camera wire keys, the state routing,
+and the deployable action width — the way OpenPI's `serve_policy.py
+--policy.config <TrainConfig>` does. It defaults to `franka_libero`; a checkpoint
+fine-tuned for another robot must name its own preset, because a mismatch
+degrades silently rather than failing. `--help` lists every registered preset
+with its slot→key mapping. See
+[Adding an embodiment](doc/adding-an-embodiment.md) for the full launch guide and
+for how to register a new robot.
+
 Start FP8 on port 8000:
 
 ```bash
--- a/crates/apxinf-py/apxinf_py.pyi
+++ b/crates/apxinf-py/apxinf_py.pyi
@@ -18,15 +18,20 @@ class Model:
         precision: str = ...,
         calibration: str | None = ...,
         tactics: str | None = ...,
+        action_horizon: int | None = ...,
         num_views: int | None = ...,
+        sampling_seed: int = ...,
     ) -> "Model":
         """Load a checkpoint through the unified ``AutoModel`` frontend.
 
         ``device`` is ``cuda:N`` (default) or ``cpu``.
         ``precision`` is ``auto`` (default), ``fp8``, ``bf16``, or ``int8``.
+        ``action_horizon`` overrides the checkpoint's chunk length (a sequence
+        length, not a weight dimension).
         ``num_views`` serves fewer cameras than the checkpoint declares (1..=its
         own count), which is numerically equivalent to openpi padding + masking
         the absent views but skips their patch tokens.
+        ``sampling_seed`` seeds the implicit device-side noise stream.
         """
         ...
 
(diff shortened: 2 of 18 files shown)


wong-1994:
Superseded by #16 — the branch was renamed (`rebase/robot-presets-onto-upstream` → `feat/robot-embodiment-presets`), which closes a cross-fork PR rather than retargeting it. Same commits plus `doc/adding-an-embodiment.md`.

PR #29, opened 2026-08-28
feat(pi05): add observation-driven FP8 calibration

wong-1994:
## Summary

- add a native BF16 eager collector and checkpoint-bound FP8 calibration profiles for PI0.5
- expose observation-driven calibration through the Python API and `scripts/calibrate_pi05.py`
- add a public-seam validator for profile reproducibility, site coverage, BF16/FP8 action accuracy, and timing evidence
- document the generalized static calibration workflow and Thor validation results

## Validation

- full Python suite: **123 passed, 14 skipped, 3 warnings, 5 subtests passed**
- focused calibration tests: **28 passed**
- Thor SM110 CUDA release workspace build and Python wheel import: **passed**
- two independent 16-observation text-only calibrations produced byte-identical 256-site profiles

## Notes

The validated LIBERO path follows the README default and does **not** enable `--discrete-state`. The 16-observation calibration is a validated lower bound, not a claim of sample-count convergence.


zealoct (changes requested):
Reviewed at head `acd103c` in a local worktree. Overall: architecture is sound, the fail-closed posture is exactly right for quantization artifacts, and the evidence docs set a high bar. I verified the two subtle correctness questions in the PR's favor: the observer's D2H copy is safely ordered (blocking streams + null-stream `cudaMemcpy` legacy semantics), and the last-language-layer tail exclusion matches `language_layer(..., index + 1 < depth)` in the FP8 runtime. Requesting changes for issue 1 below.

**1. Checkpoint-identity asymmetry breaks sharded checkpoints (blocking).** At load time `load_registered` calls `checkpoint_identity(model_dir)`, hitting the Rust dir branch: all `*.safetensors` recursively, index.json excluded. The Python calibrator pointed at `--checkpoint *.index.json` includes the index file in the digest. Any sharded checkpoint therefore yields different `sha256:` identities and every valid profile is rejected at load with "checkpoint mismatch". Extra safetensors in the model dir diverge similarly vs the Python single-file default. Please pick one canonical rule on both sides (e.g. shards only, never the index) and add a cross-language digest test vector — or fail loudly at calibration time for sharded checkpoints.

**2. Digest file-ordering divergence (low).** Rust sorts by `PathBuf` (component-wise), Python by `sorted(str(...))` — these disagree for names like `a-b/x` vs `a/b`. Same silent-divergence class as issue 1; sort both sides on the identical canonical string.

**3. Every FP8 load hashes the full checkpoint (perf/UX).** Streaming multi-GB safetensors through scalar SHA-256 adds real startup latency on Orin/Thor eMMC. The binding is the point, but please document the cost in the README, and consider a sidecar digest cache.

**4. Double observation on fused GeGLU (low).** `gemm_bf16_geglu_fused` observes before the tactic lookup; on `Ok(None)` the fallback `gemm::bf16` observes the same activation again — harmless but doubles the D2H copy. Move the observe after the fused-path decision.

**5. Dead/duplicated code in `scripts/calibrate_pi05.py` (low).** `calibration_document()` and the `merge_records` wrapper are only reachable from tests; the zero-fixture `minimum_amax` block is duplicated between `main()` and the helper — drift hazard.

**6. Nits.** Observer doc claims thread-local isolation but `records` is a `Mutex` behind an `Rc` (`!Send`) — use `RefCell` or drop the claim. Validator hard-codes `"fp8_runtime_accepted": True` (true only by ordering); derive or comment it. `_stats` p50/p95 use floor indexing.

**7. Question.** Old-format `calibration.json` profiles are now rejected with no migration path (intentional, and right) — confirm deployed model dirs get regenerated in lockstep, since the failure only appears at load time on device.

wong-1994:
All correctness issues have been addressed on the current head (`c6e129b`):

1. **Checkpoint identity / sharded checkpoints:** fixed in `c771f92`. Python and Rust now use the same rule: resolve only shards referenced by `model.safetensors.index.json`, exclude the index itself, and prefer the indexed/single canonical checkpoint over unrelated `.safetensors` files. `554ea1a` adds coverage for unreferenced extra shards. A mismatch now emits a warning instead of blocking execution, per the intended deployment policy.
2. **File ordering:** also fixed in `c771f92`. Both sides canonicalize relative paths with `/`, require UTF-8, and sort by the canonical UTF-8 bytes. `tests/fixtures/checkpoint_identity` is consumed by both the Python and Rust digest tests.
3. **Startup hashing cost:** documented in `doc/pi05-fp8-calibration.md`. A sidecar digest cache is not included in this PR; the runtime still computes the checkpoint identity at startup.
4. **Fused GeGLU observation:** fixed and guarded by `bc25dfc`. The observer now runs only after a fused route is selected; the CUDA regression verifies `Ok(None)` observes zero times and the ordinary BF16 fallback observes exactly once.
5. **Duplicated/dead calibration helpers:** the script-local `calibration_document()` path was removed during the data-source refactor. Bootstrap minima are applied once in `Pi05CalibrationJob`; record merging/document construction live in the shared calibration module and are called directly by `CalibrationRunner`.
6. **Nits:** observer records now use `RefCell`; `fp8_runtime_accepted` is derived from coverage; p50/p95 use linear interpolation, with regression tests.
7. **Old schema:** confirmed intentional: there is no migration path. Existing deployed calibration files must be regenerated with the matching ApxInf version. The detailed calibration doc now states this explicitly. We regenerated the artifact on Thor from native LIBERO10 observations and validated both a real FP8 LIBERO rollout and the FP8 WebSocket path.

wong-1994:
Double confirmed shell commands in readme. Calibration script works as expected.

PR #30, opened 2026-08-28
feat: add native WallOSS VLA runtime

wong-1994:
## Summary

- add a native WallOSS dual-stream VLA runtime with BF16 and dynamic FP8 execution on NVIDIA Thor (`sm_110`)
- use per-token activation scales, per-output-channel weight scales, CUTLASS/cuBLASLt tactics, staged prefix KV, action flash decode, and unified CUDA Graph replay
- add the required causal hdim128 FA2 specialization and make embedding lookup capture-safe
- add `WallossPolicy`, `AutoPolicy` checkpoint detection, the PyO3 patch/action-mask bridge, and an OpenPI-compatible WebSocket launcher
- keep WallOSS and PI0.5 model implementations isolated: no cross-model imports; the shared `vla` module contains only the stable request/response runtime contract
- retain direct optimized CUDA calls where WallOSS and PI0.5 use the same performance-oriented model-layer convention

## Provenance

- reference source revision: `ca013e817fead355bfca2f989b71a0d72a641a24`
- checkpoint: `wall-oss-0.5`, `model.safetensors` size `8,331,486,672` bytes
- target: NVIDIA Thor, compute capability 11.0
- requested precisions: BF16 and native E4M3 FP8

## Accuracy on Thor

Checkpoint-supported 313-token state-string fixture, fixed input and latent:

- BF16 eager vs reference eager BF16: cosine `0.999935567`, relative L2 `0.011410438`
- dynamic FP8 eager vs reference eager BF16: cosine `0.999546587`, relative L2 `0.030160574`
- BF16 graph vs BF16 eager: byte-identical
- dynamic FP8 graph vs dynamic FP8 eager: byte-identical
- repeated BF16/FP8 eager and graph runs were deterministic

Python preprocessing was compared directly with the private reference transform on the same two 256x256 RGB inputs, state, mask, and prompt:

- token IDs: exact match
- image patches after the BF16 boundary: exact match, max-abs `0`
- active DoF mask: exact match
- produced runtime contract: patches `[648,1176]`, action mask `[10,26]`

Real-checkpoint `WallossPolicy` BF16 inference returned finite `[10,7]` actions; repeating with the same explicit noise was byte-identical.

## Performance on Thor

Release runtime measurements:

- dynamic FP8 unified graph, 227-token prefix, 30 runs: steady p50 `76.238 ms`, minimum `75.857 ms`; 30/30 outputs identical
- BF16 unified graph, 313-token state-string prefix, 20 runs: steady p50 `116.134 ms`, minimum `115.909 ms`; 20/20 outputs identical

The 227-token input is a compute-shape latency case only. Accuracy evidence uses the checkpoint-supported state-string path.

The final Python policy was also exercised with a real checkpoint on Thor. With checkpoint-local stale tactics excluded, the second BF16 call measured `141.670 ms` model time and `148.700 ms` total policy time. This is a serving smoke measurement, not a replacement for the lower-level graph baseline above.

## PI0.5 no-regression gate

The parent commit `43add1b` and this PR head were built as separate CPython 3.10 CUDA wheels and tested sequentially on the same Thor with the same checkpoint, images, prompt, explicit noise, and tactics:

- L0/L1/L2 output SHA-256 values were identical between the two wheels (bitwise-identical outputs)
- real-checkpoint L2 BF16 p50: `74.893 ms` before, `74.657 ms` after (30 samples)
- synthetic paired p50 before -> after: L0 `75.06 -> 74.76 ms`, L1 `75.43 -> 75.29 ms`, L2 `75.71 -> 75.73 ms`
- the canonical PI0.5 eager-vs-graph BF16 integrity test remained bitwise identical (`max_abs=0`, cosine effectively `1.0`)

No PI0.5 accuracy or performance regression was observed.

## Python and WebSocket validation

- `AutoPolicy.from_pretrained()` recognizes the published WallOSS checkpoint signature without claiming generic Qwen2.5-VL checkpoints
- public in-process API accepts two RGB `uint8` images, state, prompt, optional DoF mask, and optional explicit noise
- real server `/healthz`, metadata handshake, MessagePack NumPy request, and `[10,7] float32` action response passed on Thor
- the WebSocket server remains model-agnostic and is unchanged; only a thin WallOSS launcher was added
- launcher exposes `--tactics` so a database generated for the current GPU/kernel build can override a checkpoint-local database

## Final validation

- model-family boundary check passed; WallOSS and PI0.5 contain no cross-model references
- `git diff --check` passed
- `cargo check -p apxinf-py --no-default-features` passed
- targeted `rustfmt --check` passed for `crates/apxinf-py/src/lib.rs`
- Python frontend suite: 77 passed, 14 skipped
- `cargo test -p apxinf-model --no-default-features walloss`: 10 passed
- clean `APXINF_CUDA_ARCH=sm_110 cargo check -p apxinf-model --features cuda` passed on Thor
- `APXINF_CUDA_ARCH=sm_110 cargo test -p apxinf-model walloss --features cuda`: 14 passed
- real-checkpoint WallOSS Python policy and WebSocket round-trip passed on Thor
- `embedding_bf16_lookup_can_be_captured` passed
- `fa2_causal_gqa_hdim128_matches_fp32_reference` passed, cosine `0.999998136`

## Review notes

- direct CUDA fast paths are intentional for performance and follow the same convention as PI0.5
- no WallOSS/PI0.5 cross-model imports or shared network/weight components are introduced
- the checkpoint-bundled tuning database used during validation is stale for the current kernel build; the runtime correctly rejected it, and the public Python/WebSocket APIs expose an explicit tactics override


zealoct:
> ops that has been added but not called:
> - `split_gqa_qkv_bias_bf16`
> - `gqa_bf16`
> - `quantize_columns_f16_e4m3`
> - `apxinf_dynamic_prepare_fp8_gemm_bf16`
> - `apxinf_dynamic_fp8_gemm_bf16`
> - `apxinf_dynamic_rescale_rows_columns_bias_bf16`
> `transpose_e4m3` only been called to generate weight_kn that is used by above APIs, so could be removed either

wong-1994:
@zealoct fixed in `d53d8e6`.

The listed unused paths were removed end-to-end, including their CUDA adapters, FFI declarations, Rust wrappers, tests, weight/workspace state, and call-site-only dependencies:

- `split_gqa_qkv_bias_bf16`
- `gqa_bf16`
- `quantize_columns_f16_e4m3`
- `apxinf_dynamic_prepare_fp8_gemm_bf16`
- `apxinf_dynamic_fp8_gemm_bf16`
- `apxinf_dynamic_rescale_rows_columns_bias_bf16`
- `transpose_e4m3` and `weight_kn`

also audited the other operators/functions introduced by this port for production call sites, including paths referenced only by tests, and removed the remaining unused experimental/residual code. Exact-symbol checks show none of the symbols above remain. The CUDA release build and the relevant Rust/Python test suites pass.

PR #31, opened 2026-08-30
feat: add one-step with warmstart for pi05 and update docs

Kimho666:


scripts/eval_libero_warmstart.py
zealoct, line 1: merge this script into eval_libero.py as a args, eval_libero.py will soom be moved

doc/run_warmstart_with_onestep.md
```diff
@@ -0,0 +1,252 @@
...
+  - adds `--flow-start-time`
+  - returns `normalized_actions` from the in-process backend
+  - constructs shifted/tail-repeated/blended noise before each cached replan
+
+## Build
+
+The CUDA binding must be rebuilt after changing the Rust runtime:
+
+```bash
+cd /home/daigroup/projects/ApxInf
```
zealoct, line 77: should we include this machine specific path?

zealoct (changes requested):

doc/run_warmstart_with_onestep.md
Kimho666: no

scripts/eval_libero_warmstart.py
Kimho666: Done 8893073

zealoct (changes requested):
Reviewed at head `8893073`. The core change is solid: the `α·shifted + (1−α)·ε` blend at `t = 1−α` matches the linear-flow interpolant exactly, all three runtimes (BF16/FP8/INT8) get the schedule change consistently with no missed `dt`/time-embedding sites, binding overrides are applied before validation and model load, and the doc's real LIBERO-10 numbers (92% warm 1-step vs 93% cold 10-step, 90% cold 1-step) make a convincing case.

Requesting changes on four items:

1. **`--warm-start-alpha 1.0` error message.** α=1 derives `flow_start_time = 0`, which trips `parser.error("--flow-start-time must be in (0, 1]")` — confusing, since the user never passed that flag. Please add a dedicated message pointing at `--warm-start-alpha` needing to be `< 1`.

2. **Doc filename typo.** `doc/run_warmstart_with_onstep.md` — "onstep" should be "onestep". Also the README line "please refer ApxInf/doc/run_warmstart_with_onstep.md" should read "refer to `doc/run_warmstart_with_onestep.md`" (repo-relative path). Cheap to fix now, annoying after merge.

3. **Missing negative-path test.** `config.rs` gains the `(0, 1]` validation for `flow_start_time` but no test exercises rejection (`0.0`, `1.5`, NaN). One `assert!(cfg.validate().is_err())` case would lock the contract in.

4. **Partial flow without warm start is a silent quality trap.** Passing `--flow-start-time 0.5` alone (without `--warm-start`) runs cold-start partial flow from pure noise with no indication anything is off. Please either warn when `flow_start_time < 1` while warm start is disabled, or call this out in the flag's help text.

One non-blocking observation, no change requested: since `flow_start_time` is a load-time constant, the first inference of each episode (no cache yet) integrates from `t = 1−α` starting from pure standard normal — a distribution mismatch that the eval shows is tolerable in practice, but worth a sentence in the doc.

PR #35, opened 2026-09-01
feat(cuda): add operator-driven GEMM tactic autotuning

ChenxiHe231:
## Summary

Add a model-independent GEMM tactic selection and autotuning system for CUDA.

Tactics are now selected by the physical GEMM contract rather than by model,
layer, or executor name. This allows compatible tactics to be reused across
PI0.5 and future models such as GR00T.

## What changed

- Add runtime GEMM tuning sessions with `Inference` and `AutoTune` modes.
- Add exact-key and shape-bucket tactic lookup.
- Benchmark and validate candidates using real inference operands.
- Support BF16, FP8, W8A8, and fused GEMM paths.
- Cache prepared GEMM plans for graph-safe execution.
- Persist winners atomically to hardware-specific `tactics.json` files.
- Write candidate measurements to `tuning_report.json`.
- Add provider-local compatibility through `implementation_version`.
- Expose `autotune=True` in the Python API.
- Add `--autotune` to the benchmark and websocket server.
- Consolidate tactic databases under:

  `configs/tuning/<vendor>/<device-family>-sm<version>/`

## Runtime behavior

The selection order is:

1. Exact GEMM key
2. Compatible shape bucket
3. Safe provider default

When autotuning is enabled, an exact-key miss is tuned using the first real
request. Candidates are checked against a safe reference before the fastest
valid implementation is persisted.

Autotuning is disabled by default. Tuning or persisted-tactic failures fall
back to the safe implementation.

## Usage

```python
policy = AutoPolicy.from_pretrained(
    "<model-path>",
    precision="bf16",
    autotune=True,
)

### CODE REVIEW

**Comment by ChenxiHe231** [comment]
## test 1: rust benchmark with autotune
### command
```
target/release/examples/pi05_bench random \
  --dtype fp8 \
  --calibration uniform:1 \
  --views 2 \
  --action-horizon 10 \
  --token-count 10 \
  --iterations 200 \
  --autotune
```
### result:

```
"latency_ms": {
    "max": 42.517692000000004,
    "mean": 41.534838729999954,
    "min": 41.22899,
    "p50": 41.479315,
    "p95": 41.919256000000004,
    "standard_deviation": 0.2084360782747485
  },
```
## test 2: rust benchmark without autotune
### command
```
target/release/examples/pi05_bench random \
  --dtype fp8 \
  --calibration uniform:1 \
  --views 2 \
  --action-horizon 10 \
  --token-count 10 \
  --iterations 200 \
```
### result:

```
"latency_ms": {
    "max": 42.517692000000004,
    "mean": 41.534838729999954,
    "min": 41.22899,
    "p50": 41.479315,
    "p95": 41.919256000000004,
    "standard_deviation": 0.2084360782747485
  },
```
## test 3: python benchmark with autotune
### command
```
python scripts/bench_pi05.py \
  --random-weights \
  --precision fp8 \
  --calibration uniform:1 \
  --layer l1 \
  --views 2 \
  --token-count 10 \
  --action-horizon 10 \
  --num-flow-steps 10 \
  --warmup 10 \
  --samples 100 \
  --autotune
```

### result
```
pi05 in-process latency  |  fp8  synthetic  H=10 Dmodel=32 views=2 T=10
-----------------------------------------------------------------------
layer                 p50      p95     mean      std      min      max
L1_rust             41.65    42.14    41.72     0.28    41.38    43.13`
```

## test 4: python benchmark without autotune
### command
```
python scripts/bench_pi05.py \
  --random-weights \
  --precision fp8 \
  --calibration uniform:1 \
  --layer l1 \
  --views 2 \
  --token-count 10 \
  --action-horizon 10 \
  --num-flow-steps 10 \
  --warmup 10 \
  --samples 100 \
```

### result
```
pi05 in-process latency  |  fp8  synthetic  H=10 Dmodel=32 views=2 T=10
-----------------------------------------------------------------------
layer                 p50      p95     mean      std      min      max
L1_rust             41.69    41.89    41.68     0.13    41.29    41.96
```

zealoct (changes requested):
# PR #35 Review — Operator-driven GEMM autotuning

**Branch:** `review/pr-35` (worktree `apxinf-pr-35`), 4 commits on top of `6390d57`
**Commits:** `a5bd89c` split tactic identities from store → `737d630` runtime tuning session → `4254aff` operator-driven GEMM autotuning → `5a38876` python entry points
**Scale:** ~4.5k lines of Rust/CUDA changes + regenerated `configs/tuning/nvidia/*` databases (bulk of the +21k)
**Review method:** full read of new `tuning/` + `gemm/plan|providers` modules; diff review of fp8/bf16/w8a8/fused kernels, cublaslt adapter, model runtime, py bindings, scripts. Not compile-tested (no cargo/CUDA in this environment). `python -m unittest tests.test_tactics`: **6/6 pass**.

## Verdict: **Request changes** (minors 1, 2, 6, 7, 8 below; majors tracked as follow-up issues)

This is a well-designed rework: the process-global `OnceLock` tactic store is replaced by a per-context `TuningSession` with generation-stamped plan invalidation, online autotune runs from real request operands (never synthetic workspace-prep traversals, never during capture), winners are verified numerically against an independent reference before selection, and persistence is lock-protected + atomic-rename. Tests are thoughtful, docs were actually updated, and legacy DB migration is careful (e.g. preserving `cublaslt_custom_bias` as `bias_residual`). The adapter change separating per-epilogue rank/custom maps fixes a real pre-existing collision (bias plans previously consulted the plain-GEMM `g_cublaslt_ranks`).

Two majors (M1, M2) are recorded here for reference — they will be tracked as separate issues and fixed in a future PR. The requested pre-merge changes are minors 1, 2, 6, 7, and 8.

---

## Major (tracked as follow-up issues, not blocking this PR)

### M1. `merge_record_atomic` destructively prunes the on-disk database
`crates/apxinf-cuda/src/tuning/db.rs:150-173` — the atomic merge re-reads the file through `build_store(caps, versions)`, which **filters out** records whose cuda/cublas major.minor or `implementation_version` don't match the *current* runtime, then rewrites the file with only the survivors. Consequences:

- One autotune run under an upgraded CUDA/cuBLAS erases every previously valid cuBLASLt record from the shared hardware DB (they aren't stale for hosts still on the old stack, and these files are committed to the repo).
- Two hosts sharing one hardware dir but differing in cublas minor version will ping-pong-wipe each other's records.

The comment on `publish_gemm_persisted` says the locked re-read exists so "concurrent model processes do not discard one another's newly discovered exact keys" — but version-incompatible records *are* discarded. Suggest merging at the parsed-record level (carry incompatible records through verbatim) or at minimum documenting that autotune intentionally regenerates the file for the current stack.

### M2. Shared plan cache vs `thread_local` native plan state
`GemmPlanCache` lives on `CudaContext` (shared across threads), but all cuBLASLt rank/custom/plan maps in `cublaslt_adapter.cu` are `thread_local`. If a plan is resolved (and natively prepared) on thread A, a later lookup from thread B hits the Rust cache and **skips native prepare entirely** — B's adapter then lazily builds a default rank-0 plan, silently diverging from the persisted/tuned tactic. The old code re-ran prepare on the launch path per thread (gated by `may_prepare_native_resources`), so each thread self-healed.

If the contract is "one context = one execution thread", this is fine — but that contract is now load-bearing and undocumented. Please confirm/document it, or key the plan cache per thread.

## Minor

1. **Native prepare during stream capture on a cache miss** — `plan.rs:118` (`prepare_and_cache`) calls `providers::prepare` unconditionally; the old code gated native prepare with `may_prepare_native_resources()`. The stabilization loop in `runtime.rs:725-750` should make a miss-during-capture impossible, but if it ever happens, cuBLASLt descriptor creation would run inside capture. A defensive `may_prepare_native_resources()` check with a loud error would make that failure diagnosable.
2. **Silent record filtering = quiet perf regression** — `db.rs:build_records` drops version-incompatible records with no log; an inference-mode load on a mismatched stack yields an empty store and fallback tactics with no signal (the tests explicitly bless this: `treats_cuda_and_cublas_versions_as_provenance`). One warning line when N records are filtered would save a debugging session. Related: FP8 real-checkpoint loads without any tactics DB used to hard-error and now silently run fallback (behavior change worth a release note).
3. **Quick-validated bucket winners persist noisy timings and stop exploration** — `engine.rs:80-104`: a bucket tactic that passes the 1-warmup/3-iteration check becomes the *exact* winner immediately: (a) its 3-iteration timing is persisted and later participates in `is_faster` comparisons against careful 25-iteration measurements; (b) no other candidate (not even the vendor default) is ever compared, so a valid-but-slow bucket tactic is locked in permanently. Consider re-measuring the winner with the full budget before persisting, or flagging quick-path records.
4. **Fused/plain key aliasing** — fused GeGLU tactics share the exact tuning key with the plain GEMM of the same `(op, m, n, k, epilogue=None)`; fused-ness lives only in the tactic backend. Online autotune never proposes fused candidates (`providers::candidates` excludes them), so on a fresh device fusion is only reachable via shipped DBs — and an online-tuned plain winner can replace a shipped fused record through `is_faster`, comparing timings of two different physical ops. Pre-existing key design, but persistence makes it stickier; worth a comment or an epilogue-like discriminator later.
5. **`DatabaseLock` never reclaims stale locks** — `db.rs:365-402`: a crashed process leaves the `.lock` file; subsequent autotunes spin 5 s then hard-error. The PID is written but never checked. Low priority (autotune-only path), but a stale-lock takeover would be cheap.
6. **Orin/RTX4090 tactics files lack the v1 header** — `configs/tuning/nvidia/{orin-sm87,rtx4090-sm89}/tactics.json` have no `schema`/`sm` fields (unlike both Thor files), so they parse via the legacy fallback and skip the database-wide SM hard-boundary check — contradicting `configs/tuning/README.md`. Regenerate for consistency.
7. **`pi05_bench --tactics` typo'd path is now silently ignored** — `examples/pi05_bench.rs:914` maps a missing explicit path to `None` (fallback tactics, claimed as production routing) where the old code errored; `auto.rs:287-289` handles the same case correctly — the bench should match.
8. **Synthetic-load guard lost in `auto.rs`** — `configure_cuda_tuning` (auto.rs:283-285) resolves `model_root` to `"."` for synthetic loads, so a stray `./tactics.json` in CWD gets picked up — contradicting the invariant still documented at `vla_runtime.rs:652`. Gate on `options.synthetic.is_none()` or update the comment. (Also generally: the default `configs/tuning` root is CWD-relative — autotune persistence location silently depends on where the process started.)
9. **Positional-API break in pyo3 signatures** — `apxinf-py/src/lib.rs:256,332` inserts `autotune` mid-signature in `Model.load`/`Model.random`, shifting positional slots for external callers. In-repo callers are keyword-only; consider trailing keyword-only args.
10. **Python SM→dir map conflates device families** — `_tactics.py:9-14` maps every SM89 device to `rtx4090-sm89`, while Rust `hardware_directory_name` would write e.g. `rtx4080-sm89`. Autotune on an unlisted same-SM device steers records into another family's directory (harmless for lookup correctness thanks to device fingerprints, but muddies the naming contract).

## Nits

- `engine.rs:91`: `.min(3).max(1)` → `.clamp(1, 3)`; the `.max(1)` is redundant given the constructor rejects 0.
- `context.rs:tuning()` panics (`expect`) on a poisoned lock while sibling paths return `Result`.
- `apxinf-py/tests/conftest.py:11` and `pi05.py:247` docstrings still describe tactics as precision-specific; `_tactics.py:51` keeps an unused `precision` param without comment; the new `autotune` kwarg is undocumented in the policy docstring.
- `bench_pi05.py:325` / websocket server: with `--autotune` the "using … tactics" message can print a not-yet-existing path.
- The 25-line generation-stabilization loop is copy-pasted into `runtime.rs`, `bf16_runtime.rs`, and `int8_runtime.rs` (only `runtime.rs` keeps the explanatory comment) — a shared helper would help.
- `examples/pi05_integrity_probe.rs:82` still uses `install_tuning_db` instead of the new `configure_tuning` plumbing (works, but inconsistent).

## Verified clean

- `CublasLtCustomConfig` bit-packing decode matches its documented layout (tile[9:0], custom[12:10], cluster[18:13], stages[24:19]; 25-bit mask) and `validate_tactic` round-trips it.
- FFI signature change (`apxinf_static_set_cublaslt_fp8_gemm_bias_custom` gains `epilogue` after `k`) matches on both the Rust and `.cu` sides; the new `apxinf_static_set_cublaslt_fp8_fused_heuristic` routes bias/gelu/residual to the correct newly-separated maps.
- No stale references to deleted `schedule.rs` items, deleted examples (`pi05_bf16_tune`, `pi05_cutlass_tune` — Cargo.toml cleaned too), or deleted `configs/pi05/*_tactics.json` (remaining mentions are intentional `imported_from` provenance in the tuning reports).
- Autotune correctness gating is sound: independent references (dequantized-FP16 cuBLAS for FP8, vendor path for BF16/W8A8/fused), finite-check + relative-L2 + cosine thresholds, invalid candidates retained in reports but excluded from selection; failed autotune degrades to fallback with a log instead of failing the request.
- Graph-capture safety: tuning requires `may_prepare_native_resources() && !is_preparing_workspace()`, the real-request eager pre-tune runs before capture, prepared graphs are generation-stamped and rebuilt on change, and the capture path re-checks stability (bounded loop with a clear error).
- Rust↔Python plumbing (`autotune` → `LoadOptions.autotune`, `allow_missing=bool(autotune)`) is consistent across `pi05.py`, `bench_pi05.py`, and the websocket server; `_tactics.py` directory names match the on-disk layout and Rust's writer for the shipped devices.

## Requested changes before merge

Minors **1** (defensive gate on native prepare during capture), **2** (warn when records are filtered / FP8 silent-fallback note), **6** (regenerate orin-sm87 and rtx4090-sm89 headers to the v1 schema), **7** (error on a missing explicit `--tactics` path in `pi05_bench`), and **8** (restore the synthetic-load guard in `auto.rs`). All are small and touch code introduced or regressed by this PR.

M1 and M2 are recorded as separate issues and will be addressed in a future PR. The remaining minors and nits can trail as follow-ups.

PR #41, opened 2026-09-02
feat(tuning): autotune fused GeGLU implementations

ChenxiHe231:
## Summary

Extend GEMM autotuning to cover complete GeGLU operators and specialized implementations.

## Changes

- Add `GeGlu` semantics to `GemmTuningKey`.
- Enumerate custom, split and dual-GEMM fused candidates.
- Validate and benchmark the final GeGLU output.
- Remove the requirement for an existing fused Exact tactic.
- Cache the winner as a `PreparedGemmPlan` before CUDA Graph capture.
- Migrate legacy fused tactic records to the new GeGLU key.
- Fall back safely to ordinary GEMM + GeGLU.

## Result

Thor FP8 PI0.5, starting from an empty tactic database:

- Autotune discovers fused/custom implementations.
- Restarted inference automatically reuses the generated tactics.
- P50: **41.416 ms**
- Production database P50: **41.607 ms**
- Eager and CUDA Graph outputs are bitwise equal.

## Validation

- 97 `apxinf-cuda` tests passed.
- 5 CUDA architecture tests passed.
- Rust formatting checks passed.

### DIFF
--- a/configs/tuning/nvidia/thor-sm110/tactics.json
+++ b/configs/tuning/nvidia/thor-sm110/tactics.json
@@ -433,7 +433,7 @@
     {
       "key": {
         "activation_dtype": "bf16",
-        "epilogue": "none",
+        "epilogue": "geglu",
         "k": 2048,
         "layout": "row_major",
         "m": 522,
@@ -1742,10 +1742,10 @@
         "k": 2048,
         "activation_dtype": "f8e4m3",
         "weight_dtype": "f8e4m3",
-        "output_dtype": "f16",
+        "output_dtype": "f8e4m3",
         "layout": "row_major",
         "scale_mode": "per_tensor",
-        "epilogue": "none",
+        "epilogue": "geglu",
         "workspace_limit": 18446744073709551615
       },
       "tactic": {
@@ -1842,10 +1842,10 @@
         "k": 2048,
         "activation_dtype": "f8e4m3",
         "weight_dtype": "f8e4m3",
-        "output_dtype": "f16",
+        "output_dtype": "f8e4m3",
         "layout": "row_major",
         "scale_mode": "per_tensor",
-        "epilogue": "none",
+        "epilogue": "geglu",
         "workspace_limit": 18446744073709551615
       },
       "tactic": {
--- a/configs/tuning/nvidia/thor-sm110/tuning_report.json
+++ b/configs/tuning/nvidia/thor-sm110/tuning_report.json
@@ -792,7 +792,7 @@
     {
       "key": {
         "activation_dtype": "bf16",
-        "epilogue": "none",
+        "epilogue": "geglu",
         "k": 2048,
         "layout": "row_major",
         "m": 522,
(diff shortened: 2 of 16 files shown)


zealoct (changes requested):
## Review: approve with changes

Nice restructuring — GeGLU as a first-class operator with `GemmThenGeGlu` as the always-available reference tactic is the right shape, and the legacy-record migration is correct (checked the 3 hand-edited Thor records against what the loader would produce).

### Must fix
1. **`crates/apxinf-cuda/src/tests/bf16.rs::bf16_geglu_fallback_observes_activation_once` now fails.** It asserts `bf16_geglu_fused(..., false, None, None)` returns `None` and doesn't observe; with this PR it always returns `Some` (decomposed path) and observes once. Update the test to the new contract.

### Should fix
2. **Spurious "rejected persisted GEMM tactic" on Thor prefill shapes.** `m_bucket(522) == m_bucket(789) == 1024`, and the Thor DB only has GeGLU records for m=522/533 (shape-pinned dual kernels). The bf16 m=789 / fp8 m=778 GeGLU keys bucket to those, get rejected in `cutlass::prepare`, and eprintln on every fresh process. Add explicit `gemm_then_geglu` records for those keys, or exclude shape-pinned backends from bucket derivation.
3. **Plain plan + `gate` buffer are resolved/allocated even for fused winners.** `gemm_*_geglu_fused` always runs `resolve_*_plan(plain_key)` (a full plain autotune in AUTO_TUNE mode) and allocates `gate = m × full_n` (~34 MB at m=522) even when the winner is `cutlass_*_dual_geglu`, which uses neither. Make both lazy.
4. **Undocumented change to plain-GEMM tuning on SM110.** `cublaslt::candidates` now injects `CublasLtCustom` / `CublasLtCustomSplitSerial` / `CublasLtCustomBias` config IDs for all plain/bias keys, and BF16 autotune no longer filters to `Vendor|CublasLt`. Fine if intentional, but please call it out in the PR description.
5. **`cublaslt::prepare` accepts GeGLU keys for `CublasLt`/`CublasLtCustom`/`CublasLtCustomSplitSerial`,** which no GeGLU launcher can execute. A bad DB record passes `prepare_and_cache_with`, gets cached as Exact, then fails at first launch. Drop the `| Epilogue::GeGlu` additions on those arms so it fails closed at prepare.

### Nits
- `gemm_bf16_geglu_fused` can no longer return `Ok(None)`; the `None =>` arms in `bf16_executor.rs` are dead.
- `Bf16DualGeGluMode` / `Fp8DualGeGluMode` + parsers/routes are now `#[cfg(test)]`-only just to keep old tests compiling — delete them and their tests (model layer has its own copies).
- FP8 interleaved-only default is unconditionally `CutlassFp8DualGeGlu`; BF16 twin checks `sm == 110` and errors clearly. Align.
- Test gaps: BF16 case for `normalize_legacy_geglu_key`, `parse_legacy_tactics` GeGLU branch.

PR #44, opened 2026-09-07
Add native tokenizers and Torch-free WallOSS preprocessing

wong-1994:
## Summary

Built-in WallOSS inference no longer requires Python Torch or Transformers, and both WallOSS and PI0.5 now execute tokenization through `apxinf-tokenizer` and thin PyO3 bindings. WallOSS also gains a graph-capturable native RGB preprocessing path while preserving custom Python processors.

- Load legacy tensor-only `.pth` normalizers with a restricted Python reader; replace Transformers image processing with Pillow/NumPy compatibility processing and Rust/CUDA tensor preprocessing.
- Use the native HF `tokenizer.json` backend for WallOSS and a statically linked C++ SentencePiece backend through Rust for PI0.5. Remove the Python `walloss` and `tokenizer` extras and production dependencies on Python `tokenizers`/`sentencepiece`; keep SentencePiece as a test reference. Built-in inference still requires `apxinf-py`; NumPy, Pillow and PyYAML remain base dependencies.
- Move checkpoint-fixed RGB normalization, temporal duplication, patchification, spatial-merge ordering and BF16 conversion into Rust/CUDA. Advertise the RGB route through `VlaContract`; custom processors and legacy checkpoints without processor metadata retain the canonical patch route.
- Document model-neutral inference lifecycle stages, inter-stage contracts, instance/session/request lifetimes and directory ownership in `doc/model-lifecycle.md`. Keep model-specific state/token/output differences in the Python usage guide rather than requiring one architecture document per model. Correct stale tokenizer and PI0.5 state-injection documentation, and update the root build guide with an RPATH-safe wheel command.

## Scope and compatibility

Prompt construction, state normalization/discretization orchestration, image-token expansion, policy masks and action postprocessing remain Python family-policy responsibilities. This PR does not introduce a unified Prompt Builder or a standalone Rust policy. Those are follow-up changes, including canonical normalizer sidecars and Python delegation to native family builders.

PI0.5 and WallOSS retain their distinct state binning, BOS/newline handling and multimodal token rules. Custom Python processors remain supported. The optional LeRobot adapter retains its separate Torch requirement.

Review fixes bound and cache aliased storages in the restricted reader, prevent implicit native routing for custom processors, validate processor JSON types and derived dimensions before FFI, and use model-neutral CUDA/FFI/operator names.

### Rechecked during documentation closeout (2026-09-08)

- Thor Python suite: **183 passed, 12 skipped, 5 subtests passed**. The tested snapshot's 61 Python files match the PR source except for one non-executable dependency comment in `walloss.py`.
- Real tokenizer assets: **10 SentencePiece samples** checked with and without BOS and **12 HF samples**, including Unicode, whitespace and added tokens, plus token-ID lookups: exact parity with the Python reference backends.
- Skips: 2 optional WebSocket tests, 9 optional Torch/LeRobot tests, and 1 checkpoint-gated PI0.5 test. Explicitly enabling that checkpoint test still skips because the available tokenizer-test extension was built **without CUDA**. These counts are not a fresh full-policy GPU validation.
- Documentation diff whitespace and local link targets checked. No runtime code changed in the closeout commit.

### Earlier GPU evidence recorded for the native preprocessing change

- Model-family boundary check passed; Rust/CUDA graph tests on Thor: **4 passed** (eager/captured parity, repeated replay, changed-input propagation, prepared-mode/layout validation); derived-dimension safety tests: **3 passed**.
- Full WallOSS checkpoint: eager RGB, captured RGB, repeated replay and golden BF16-patch routes produced bitwise-equal outputs, shape `[10, 26]`, all finite.
- Changed RGB propagated to output (`max_abs=0.056640625`, `relative_l2=0.02985570294159166`).

These earlier GPU results were not rerun during documentation closeout. Private checkpoints and generated golden evidence are not included in the repository.

### Two-model CUDA end-to-end regression (2026-09-08)

- Compared PR head `bf5ab02` with its exact parent `d8f19ce` on NVIDIA Thor, BF16, using the same checkpoint assets, fixed RGB/state/instruction inputs and noise. PI0.5 used two views, horizon 10, 10 flow steps, state injection and float64 normalization; WallOSS used its checkpoint profile. Both used the same unchanged tactic selection and normal prepared-execution entry points.
- Three deterministic synthetic observations per model, two process rounds in reversed baseline/candidate order, 3 warmups and 12 timed repetitions per case: **72 timed full-policy samples per variant/model**. CUDA was explicitly synchronized around full-policy timing.
- **Byte-identical token IDs, normalized actions and final actions across both variants and both rounds; maximum absolute error 0.** Every timed repetition checked stable, finite final actions.
- Candidate full-policy execution imported neither Torch nor Transformers.
- Reused 21 verified unchanged CUDA objects from the parent build after comparing 1,599 CUDA source/header inputs. Recompiled both adapters affected by the changed preprocessing header and all changed Rust/PyO3 code; CUDA-enabled release build took **29.24 seconds**. Object reuse used a private build-script overlay, not a product-code change. The baseline installed extension matched its original build byte for byte.

| Model | Baseline full-policy median / p95 | PR full-policy median / p95 | Median change |
| --- | --- | --- | --- |
| PI0.5 | 82.92 / 101.95 ms | 83.14 / 102.11 ms | +0.27% |
| WallOSS | 122.63 / 139.41 ms | 115.99 / 119.49 ms | -5.41% |

CPU preprocessing median: PI0.5 **2.487 → 2.565 ms**, WallOSS **8.113 → 4.296 ms**. Model-call median: PI0.5 **79.923 → 80.133 ms**, WallOSS **110.873 → 111.218 ms**. CPU preprocessing was sampled separately; medians are not additive. No material end-to-end slowdown was observed; WallOSS's reduced host preprocessing lowered total latency in this sample.

The prior CPU-only test-environment limitation was addressed with a new CUDA-enabled extension loaded directly from its release artifact. Reproduction scripts, raw outputs/timing vectors, checkpoint hashes and binary/object provenance are archived privately. No runtime source change was required.

## Evidence limits

The paired CUDA check closes the previously outstanding full-policy parity/latency check for these two BF16 configurations. It uses a finite synthetic corpus on a shared Thor with two other resident processes observed idle, without locked clocks or dedicated-host isolation. The small PI0.5 timing difference is not evidence of a material regression, and these measurements are not statistical equivalence guarantees. Cold loading/build/capture time is excluded. No broad closed-loop LIBERO success-rate evaluation, other precisions/devices, exhaustive checkpoint coverage, or fresh wheel-installation validation is claimed. Private checkpoints and generated tensor evidence remain outside the repository.


zealoct:
does the performance decline on pi05 expected? or is it statistical error?

PR #62, opened 2026-09-12
feat(pi0-fast): BF16 inference in a persistent arena and doc the module (70% SR on LIBERO-10, 32.2 ms per token)

Kimho666:


zealoct:
@wong-1994 check the func of python layer and scripts, add an issue to port to apxinf-robo if needed

python/apxinf/pyproject.toml
```diff
@@ -20,6 +20,19 @@ dependencies = [
...
+# π0-FAST needs no torch: both tokenizers (PaliGemma for the prompt and id space,
+# FAST for the BPE detokenization) load through `tokenizers` — the engine behind
+# HF's fast tokenizers — straight from the assets the checkpoint names, and the
+# DCT comes from scipy. The ApxInf binding is still the only runtime.
+pi0fast = ["apxinf-py>=0.1.0", "tokenizers>=0.15", "scipy>=1.9"]
+# The lerobot adapter (apxinf.adapters.lerobot) returns torch tensors, and its
+# build_inference_frame delegates to lerobot's own build_dataset_frame. lerobot
+# itself is deliberately *not* pinned here: the adapter is built against a moving
+# API, and users bring their own already-installed version.
+lerobot = ["torch>=2.0"]
```
wong-1994: This extra is not necessary.

python/apxinf/pyproject.toml
```diff
@@ -20,6 +20,19 @@ dependencies = [
 [project.optional-dependencies]
 # A live Pi05Policy binds the PyO3 model handle (Phase 1).
 policy = ["apxinf-py>=0.1.0"]
+# π0-FAST needs no torch: both tokenizers (PaliGemma for the prompt and id space,
+# FAST for the BPE detokenization) load through `tokenizers` — the engine behind
+# HF's fast tokenizers — straight from the assets the checkpoint names, and the
+# DCT comes from scipy. The ApxInf binding is still the only runtime.
+pi0fast = ["apxinf-py>=0.1.0", "tokenizers>=0.15", "scipy>=1.9"]
```
wong-1994: By design, extras should not be added in terms of model. We do have our own implementation of tokenizer, and scipy is only used for idct. Try using our native version.

python/apxinf/apxinf/processors/resize.py
```diff
@@ -15,7 +24,7 @@
 
 from .base import ProcessorStep
 
-__all__ = ["ParseImage", "ResizeWithPad"]
+__all__ = ["ParseImage", "ResizeWithPad", "TorchResizeWithPad"]
```
wong-1994: Torch is not included. This class name should not contain "torch".

python/apxinf/apxinf/processors/resize.py
```diff
@@ -74,3 +83,82 @@ def __call__(self, image: np.ndarray) -> np.ndarray:
         pad_width = max(0, int((size - resized_width) / 2))
         canvas.paste(resized, (pad_width, pad_height))
         return np.asarray(canvas)
+
+
+def _torch_interpolate_taps(in_length: int, out_length: int):
```
wong-1994: Same. Torch prefix is kind of misleading.

python/apxinf/apxinf/processors/resize.py
```diff
@@ -74,3 +83,82 @@ def __call__(self, image: np.ndarray) -> np.ndarray:
...
+    return top * (np.float32(1.0) - vertical) + bottom * vertical
+
+
+class TorchResizeWithPad(ProcessorStep):
+    """Letterbox-resize a ``uint8`` HWC image the way LeRobot does.
+
+    Same geometry as :class:`ResizeWithPad` — scale the longer side to ``size``
+    and center the result on a zero canvas — but the interpolation is
+    ``F.interpolate(mode="bilinear", align_corners=False)``, which is what
+    ``lerobot.policies.pi0_fast.modeling_pi0_fast.resize_with_pad_torch`` calls
```
wong-1994: What about a pi0_fast model with openPi checkpoint? This resize method might introduce error.

python/apxinf/pyproject.toml
```diff
@@ -20,6 +20,19 @@ dependencies = [
...
+# The lerobot adapter (apxinf.adapters.lerobot) returns torch tensors, and its
+# build_inference_frame delegates to lerobot's own build_dataset_frame. lerobot
+# itself is deliberately *not* pinned here: the adapter is built against a moving
+# API, and users bring their own already-installed version.
+lerobot = ["torch>=2.0"]
+# The π0-FAST tests read a miniature checkpoint with the same tokenizer/scipy
+# stack the policy itself needs, so both are part of the test extra.
+test = ["pytest>=7", "sentencepiece>=0.1.99", "tokenizers>=0.15", "scipy>=1.9"]
```
wong-1994: Line39 has another test extra. Seems like bug caused by conflict resolve.

Kimho666:
@zealoct @wong-1994 
Review follow-up on the π0-FAST port. The frontend now depends on the ApxInf
  binding and numpy only, nothing in the package is named after a model, and the
  torch/scipy/HF-tokenizer shims the first cut pulled in are gone.

  Extras
  - Drop the `pi0fast` extra (apxinf-py + tokenizers + scipy) and the `lerobot`
    extra (torch); `apxinf/adapters/` is empty, so the latter served nothing.
  - Merge the two `test` extras a conflict resolution had left behind into one,
    `test = ["pytest>=7", "sentencepiece>=0.1.99"]`. The extra holds reference
    implementations only - `sentencepiece` is the differential reference for the
    native tokenizer - never a runtime dependency.

  Tokenizers
  - The PaliGemma prompt/id-space tokenizer and the FAST BPE detokenizer both load
    through `apxinf_py.HfTokenizer`, so the HF `tokenizers` wheel is no longer
    needed; `encode()` likewise goes through the binding.
  - Read the base vocabulary from the asset itself (`model.vocab`, or
    `model.tokens` for Unigram) rather than from the binding, whose `vocab_size`
    counts added tokens: PaliGemma carries `<image>` just past the base
    vocabulary, and an off-by-one there shifts every action id in the stream.

  Inverse DCT
  - Replace `scipy.fft.idct(x, axis=0, norm="ortho")` with a cached numpy basis
    multiply. The DC weight `sqrt(0.5)` belongs to the coefficient index
    (column 0), not to the first output sample.
  - Cross-checked against scipy once, offline: max |diff| 1e-13 over (10, 7),
    (10, 2) and (16, 32) blocks, so dropping the dependency moves no numbers.

  Resize
  - Rename `TorchResizeWithPad` -> `ResizeWithPadNoAntialias` and `_torch_*` ->
    `_bilinear_*`. The step is a two-tap half-pixel kernel transcribed from
    `F.interpolate(align_corners=False)`; its arithmetic never needed torch, and
    the old name also implied a dependency the package does not have.
  - Document the three references a π0-FAST checkpoint can come from: PIL
    BILINEAR (OpenPI client, π0.5's pipeline), the two-tap kernel (LeRobot's
    `resize_with_pad_torch` - the default here), and OpenPI's JAX
    `jax.image.resize(..., LINEAR)`, which antialiases. The last one cannot be
    inferred from the weights, so such a caller passes `image_pipeline=`
    explicitly. The kernels disagree by up to ~52/255 on random content and
    ~21/255 on LIBERO frames - enough to change which FAST tokens are emitted.

  Tests and docs
  - `_idct` is now the term-by-term orthonormal DCT-III definition plus the stub's
    coefficient scale, so the check shares no arithmetic with the policy's basis
    product; `test_orthonormal_idct_matches_the_definition_one_harmonic_at_a_time`
    probes one harmonic at a time, which is what pins the DC convention the
    transposed-basis slip hid.
  - `pytest.importorskip("tokenizers")` -> `pytest.importorskip("apxinf_py")`.
  - Root README installs the binding wheel then `pip install -e python/apxinf`,
    since there is no extra left to name; `python/apxinf/README.md` spells out
    what the `test` extra is for.

PR #72, opened 2026-09-15
perf(qwen-drive): Jetson AGX Orin sm_87 optimisation — 6.61s to 6.01s per VQA scene

qhy991:
Optimisation of Qwen-Drive-1.0-4B on **Jetson AGX Orin (sm_87, JetPack R36.5 / CUDA 12.6)**.

This is a different device and a different exercise from the RTX 4090 pilot on `codex/qwen-drive-k3-round20-20260914`. That branch states up front that "correctness differences versus the official reference are accepted; this is not a correctness qualification". Here every change is checked against the previous binary, and the ones that are not bit-exact say so and show the measurement.

## Result

Fixed four-scene VQA workload, batch 1, BF16, 64 generated tokens, one warmup and three measured requests per scene, warm median:

| | start of this work | now |
|---|---:|---:|
| per scene | 6.6088 s | **6.0143 s** |
| fixed cost (vision + prefill) | 3.073 s | **2.49 s** |
| decode | 55.5 ms/token | 55.2 ms/token |

**9.0%** overall, essentially all of it in prefill.

Decode did not move and should not have. It reads 8.41 GB of weights per token, which at 55.2 ms is **152 GB/s against a 204.8 GB/s bus** — the practical LPDDR5 roofline. Prefill is now **95.8% GPU-busy** (2294.7 ms wall against 2197.5 ms of kernel time), so the host-side headroom is gone too; what remains has to come out of kernel time.

## What changed

**GDN scan kernels.** `gdn_chunk_state` was 799 ms, 29% of prefill, for 51.6 GFLOP — about one percent of what the board can do. It was not short of bandwidth or flops, it was recomputing: both inner loops consume `v_new` only through a BF16 round trip, and each element was rounded once per row of the matrix it multiplies, 64x and 128x at the shipped shape, with an `exp2` per cell on a value that depends only on the chunk row. Hoisting that into a shared tile took it to 555 ms. A second pass tiled the accumulators so the shared operand leaves the inner loop, and the same treatment applied to `gdn_chunk_gemm`.

The tile width is a template parameter with a runtime switch, because it is not guessable and not the same for the two kernels — 8 for chunk-state, 16 for chunk-gemm, and in the chunk-gemm sweep 16 is 8% faster than the 32 that looked like the obvious ceiling. Two failed attempts are recorded in the commits: a runtime trip count spills the accumulators to local memory and nearly doubles the kernel's cost, and recomputing the row index by integer division inside the unrolled body costs more than the shared read it saves.

**Projection GEMMs reach the tuning system.** Every projection except `gate_up` went through `gemm::write_ex`, a raw cuBLAS call with no tactic lookup, no cuBLASLt plan, and invisible to autotune — `zba`, `down`, `o_proj`, the GDN out projection and the packed qkv. That is why an autotune pass over this model wrote exactly one record. The loader now stores them as `[in, out]` so the GEMM is row-major NN, the same treatment `gate_up` already had. Geometry alone is 4.7%; tuning on top is a further 2.4%. Autotune now writes 13 records where it wrote 1, and they are merged into `configs/tuning`.

**Vision position embeddings.** `compute_pos_embeds` is a pure function of a checkpoint constant and `grid_thw` but ran per request: a table readback, 12.5M scalar bilinear taps on the host, and a 25 MB upload — 41.5 ms. Cached on `VisionDeviceWeights` rather than in a process-global map, so it cannot outlive its model or be shared between two of them. This one is ported from the 4090 branch; there it is worth 4.6%, here 0.74%.

**Decode CUDA graphs, fixed.** Capture and replay produced wrong tokens and the earlier diagnosis blamed the workspace arena. Five arms on one binary overturned that: every arena arm is bit-identical and only capture-and-replay diverges. Two bugs, both host state that only advances when the Rust body runs — the conv ping-pong flip that also selects the graph, and captured graphs holding the addresses of cache buffers that `reset_state` replaces. Correct now, and worth 0.16%, so off by default; the allocation cache had already removed the launch overhead graphs exist to remove.

## Verification, and one thing worth knowing

The FA2 head-256 causal prefill kernel **does not reproduce itself**. Identical q, k and v bit for bit, and about one output element in ten thousand comes back one BF16 ULP different, which moves where generation first diverges from the reference — the gate's `first_different_index` gave 121, 121, 255 on three consecutive runs of one binary. Prefill traces localise it exactly: vision reproduces, text layers 0-2 reproduce, layer 3 does not, and inside layer 3 only the attention output differs. Under `APXINF_CUDA_SKIP_OUTPUT_ZERO=poison` there is no NaN anywhere, so nothing uninitialised is being read.

That makes a single-run token hash worthless as a regression signal, so every bit-exactness claim here is verified under `APXINF_ATTN_COMPOSED_PREFILL=1`, which routes prefill through the composed path and reproduces. It costs 4.9% and is left off by default; whether that trade is worth making is a product decision, not one made here.

All changes are bit-exact except the projection-GEMM one, which selects a different kernel and so changes the BF16 reduction order. Measured: text layers 0-2 identical, layer 3 differs at 1.8e-3 relative with a maximum absolute delta of one BF16 ULP, amplified by the residual stream to 2.6e-2 by layer 31. Nothing observable moved — the 130-token VQA probe is identical, and both trajectory modes report the same error to every digit. `APXINF_QWEN_LINEAR_TUNED=0` restores the raw path.

Gate status is unchanged throughout: vqa diverges at index 121 with token 357, direct and reasoning sit in their existing 0.0403/0.0806 pair, perception remains the declared BEV gap.

## Relative to the 4090 branch

Its other items were already here, and two measured very differently on this board: packed input projections are 10.2% there and 1.05% here; removing the unconditional diagnostic readbacks is 1.3% there and, under a controlled A/B with the same binary, about 0% here. Its `fa2_attention_splitkv_into` vision-segment scratch reuse was measured and **not** taken: the vision window is 476.9 ms wall against 453.3 ms of kernel time, so 23.6 ms of idle is the ceiling for any host-side change there, and the allocation cache already has most of it.

## Environment gates

`APXINF_CUDA_ALLOC_CACHE=1` and `APXINF_CUDA_SKIP_OUTPUT_ZERO=1` are the difference between 16.3 s and 6.0 s per scene and are **not** default; `run_perf.sh` sets neither, so numbers taken through it describe a configuration nobody runs. `APXINF_QWEN_DECODE_GRAPH`, `APXINF_ATTN_COMPOSED_PREFILL`, `APXINF_QWEN_LINEAR_TUNED`, `APXINF_QWEN_POS_CACHE`, `APXINF_GDN_CHUNK_TILE` and `APXINF_GDN_CHUNK_GEMM_TILE` are documented at their definitions.

### DIFF
--- a/crates/apxinf-model/src/lib.rs
+++ b/crates/apxinf-model/src/lib.rs
@@ -11,6 +11,7 @@ pub mod auto;
 pub mod profiling;
 pub mod pi05;
 pub mod qwen3vl;
+pub mod qwen_drive;
 pub mod vla;
 mod walloss;
 
@@ -27,6 +28,7 @@ pub use auto::{AutoModel, LoadOptions, LoadedModel, ModelPrecision, SyntheticWei
 pub use profiling::GenerationProfile;
 pub use pi05::{Pi05Config, Pi05PerformanceProfile};
 pub use qwen3vl::{GeneralQwen3VL, Qwen3VLConfig, Qwen3VLTextWeights};
+pub use qwen_drive::QwenDriveConfig;
 pub use vla::{
     Action, ImageLayout, InferenceSpec, InitialLatent, Observation,
     PreparedInference, VisionObservation, VlaContract, VlaRequest, VlaRuntime,
--- a/crates/apxinf-model/src/qwen_drive/config.rs
+++ b/crates/apxinf-model/src/qwen_drive/config.rs
@@ -1,0 +1,553 @@
+//! Qwen-Drive-1.0 configuration, parsed from the checkpoint `config.json`.
+//!
+//! The schema mirrors `configuration_qwen_drive.py` in the pinned reference
+//! (revision 28091c1532e869bc7aee91fc0aef6b3e6fd0b2e0): a Qwen3.5 hybrid VLM
+//! (`vlm_config`) plus a flow-matching planning expert (`expert_config`), with
+//! trajectory and sampler constants at the top level. Parsing here keeps the
+//! loader crate model-agnostic and matches "each model owns its config".
+
+use std::path::Path;
+
+use apxinf_core::{Error, Result};
+
+/// Text-decoder geometry of the Qwen3.5 VLM. The decoder is hybrid: layers
+/// listed as `full_attention` carry a standard GQA KV cache (these are the
+/// caches the planning expert reads), while `linear_attention` layers are
+/// gated delta net (GDN) linear-attention layers with a recurrent state.
+#[derive(Clone, Debug)]
+pub struct QwenDriveTextConfig {
+    pub hidden_size: usize,
+    pub intermediate_size: usize,
+    pub n_layers: usize,
+    pub n_heads: usize,
+    pub n_kv_heads: usize,
+    pub head_dim: usize,
+    pub vocab_size: usize,
+    pub max_position_embeddings: usize,
+    pub rms_norm_eps: f32,
+    /// Qwen3.5 emits a per-head sigmoid gate on the attention output.
+    pub attn_output_gate: bool,
+    /// Per-layer type names: `full_attention` or `linear_attention`.
+    pub layer_types: Vec<String>,
+    // GDN geometry (only meaningful for `linear_attention` layers).
+    pub linear_num_key_heads: usize,
+    pub linear_key_head_dim: usize,
+    pub linear_num_value_heads: usize,
+    pub linear_value_head_dim: usize,
+    pub linear_conv_kernel_dim: usize,
+    // Partial interleaved mRoPE.
+    pub rope_theta: f32,
+    pub partial_rotary_factor: f32,
+    pub mrope_section: [usize; 3],
+    pub mrope_interleaved: bool,
+    pub tie_word_embeddings: bool,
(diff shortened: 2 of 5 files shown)


qhy991:
Closing in favour of #78, which consolidates the three device lines onto current `main`.

This branch is the trunk of that one: all 42 commits are in it, and the Thor work was built on top of them. Its constants are now the sm80-family row of `kernels/gdn_policy.rs` rather than hard-coded defaults in the adapter, which is what let the Thor row exist beside them without either board's sweeps being overwritten.

Two things from here were re-measured on Thor and came out differently, both recorded in #78 with their numbers: the BF16 chunk-gemm tiles, worth taking on Orin's 164 KB of shared memory per SM and a regression on Thor's 228 KB, and the chunk-state inter-term tiling, 10% slower there. Both stay as they are for the sm80 family. The 1024-thread chunk-state block from `80b0ecc` is kept unconditionally as committed here, except that the new policy path falls back rather than launching when the shape cannot take it — on Thor that kernel's register demand at 1024 exceeds the budget and returns `CUDA 701`.

Orin has not been re-run since the consolidation and should be before #78 merges.

PR #51, opened 2026-09-09
feat(qwen-drive): integrate Qwen-Drive with KerSor and DSH on RTX 4090

qhy991:
## What this PR demonstrates

This draft adds the current Qwen-Drive-1.0-4B native ApxInf integration and documents the agent-driven development path used to build it. The integration is being developed with KerSor Mission-v1, DSH and the ApxInf model-porting skill, targeting BF16 inference on an NVIDIA RTX 4090.

KerSor supplies the bounded Mission planner, resumable FuturePlan DAG, transaction boundary and evidence-driven replanning. DSH supplies the Infini-AI/Kimi K3 model activation path. The ApxInf Host owns candidate materialization, GPU-broker submission, remote receipts and the final four-mode verifier.

## Integration surface

- register Qwen-Drive with ApxInf model and `AutoPolicy` dispatch;
- normalize the official checkpoint layout, including mixed-dtype tensors;
- add device-resident language, vision, expert and planner state;
- connect the public Python policy through the existing PyO3 boundary;
- add the CUDA attention and linear-attention FFI/kernel integration;
- preserve the existing Host-owned reference inputs and acceptance contract.

The detailed design and evidence record is in [`doc/kersor-qwen-drive-4090.md`](doc/kersor-qwen-drive-4090.md).

## RTX 4090 evidence

The current candidate has been built and executed on a real RTX 4090 through the exclusive GPU broker:

- the native CUDA extension compiled successfully;
- 723 model tensors loaded to the device in about 60 seconds;
- the VQA path entered inference, completed all 24 vision blocks and reached `vision_done`;
- the text prefill reached `prefill_layer k=3`;
- the fixed 300-second VQA gate expired at the following causal hdim256 path before `inference_returned`.

This proves that the model is being integrated and exercised on the target GPU. It does not yet prove full four-mode deployment acceptance. The PR remains draft for that reason.

The next KerSor revision targets the measured causal text-prefill bottleneck while retaining the working vision route. A new remote receipt is required before promoting this candidate.

## Review boundary

Please review the ApxInf integration and the separation between model code, DSH activation, Host evaluation and acceptance evidence. The companion KerSor planning changes are in [KerSor PR #117](https://github.com/qhy991/KerSor/pull/117).

The parallel Codex experiment reached native build/load and VQA diagnosis but then stopped because the Codex provider quota was exhausted. It has no new accepted ApxInf result, so this PR does not claim a second Codex-backed integration.

## Validation

[...]

PR #73, opened 2026-09-16
perf(qwen-drive): RTX 4090 KerSor + DSH/K3 performance pilot — 1.1601x over eight VQA cells

qhy991:
Qwen-Drive-1.0-4B on **RTX 4090**, integrated through KerSor and DSH/Kimi-K3, with the measured performance campaign preserved in `doc/qwen-drive-performance.md`.

This is the 4090 pilot. The Jetson AGX Orin (sm_87) work is separate and is in #72 — different device, different baseline, and a different standard of evidence; the two should be read and reviewed independently.

## Scope

A performance pilot, not a correctness qualification. The report states this in its opening line: correctness differences versus the official reference are accepted for this experiment. Reviewers should not read a passing timing verifier as deployment readiness.

## Result

The round20 candidate reaches **1.160087x** against the campaign baseline over eight 64/256-budget VQA cells, with all output lengths matched, `valid=true` and `passed=true`. The changes in that step reuse vision attention scratch and GDN prefill scratch and remove redundant GEMM output zeroing.

The campaign that led there, each step measured rather than assumed:

| step | against | ratio |
|---|---|---:|
| diagnostic readback removal | fixed original baseline | 1.013026x |
| position-embed cache | frozen baseline | 1.047829x / 1.049563x paired |
| packed input projections | frozen baseline | 1.114349x replay |
| cross-module decode buffer reuse | frozen packed-projection baseline | 1.060107x |
| prefill allocation reuse (round20) | campaign baseline | 1.160087x |

Two intermediate candidates were measured and **not** promoted: decode workspace reuse at 1.005600x and the mRoPE cache at 1.008897x, both `passed=false`. They are recorded rather than dropped, which is the useful part of the log.

## What the report already flags against itself

These are the author's own caveats, carried here so review does not have to rediscover them:

- Round20 is **one measurement batch, not independent replication**. A replay attempt on 2026-09-14 was not confirmed launched because SSH connections reset during handshake — infrastructure unknown, not a verdict on the candidate.
- **Process-global vision position caches**: not owned per model, not bounded, not keyed on weights or device context. Listed as an unresolved delivery limitation.
- **Public forward last-row cropping**: the prototype crops logits in the shared text path and so does not retain the public multi-token forward output shape. Must be scoped to generation before it can be promoted as an API-compatible implementation.
- An earlier batch recorded a 1.001804x ratio that was **invalid as a comparison** — both jobs loaded the same baseline library because the source modification time preceded Cargo's build timestamp. The Host now timestamps changed inputs and rebuilds before measuring. The earlier "failed to reproduce" interpretation was withdrawn; the affected observation remains unreplicated.
- Nsight Systems 2023.3.1 failed one profiling attempt and exported no CUDA kernel records; its partial API statistics are not valid for attribution. The timings above come from the no-profiler benchmark.

## Workload

Four fixed public VQA scenes plus 256-budget cells, batch 1, BF16, fixed generated-token counts, one warmup and three measured requests per cell, host-measured `policy.infer` latency excluding weight and input loading. No profiler attached during timing.

### DIFF
--- a/crates/apxinf-model/src/lib.rs
+++ b/crates/apxinf-model/src/lib.rs
@@ -11,6 +11,7 @@ pub mod auto;
 pub mod profiling;
 pub mod pi05;
 pub mod qwen3vl;
+pub mod qwen_drive;
 pub mod vla;
 mod walloss;
 
@@ -27,6 +28,7 @@ pub use auto::{AutoModel, LoadOptions, LoadedModel, ModelPrecision, SyntheticWei
 pub use profiling::GenerationProfile;
 pub use pi05::{Pi05Config, Pi05PerformanceProfile};
 pub use qwen3vl::{GeneralQwen3VL, Qwen3VLConfig, Qwen3VLTextWeights};
+pub use qwen_drive::QwenDriveConfig;
 pub use vla::{
     Action, ImageLayout, InferenceSpec, InitialLatent, Observation,
     PreparedInference, VisionObservation, VlaContract, VlaRequest, VlaRuntime,
--- a/crates/apxinf-model/src/qwen_drive/config.rs
+++ b/crates/apxinf-model/src/qwen_drive/config.rs
@@ -1,0 +1,553 @@
+//! Qwen-Drive-1.0 configuration, parsed from the checkpoint `config.json`.
+//!
+//! The schema mirrors `configuration_qwen_drive.py` in the pinned reference
+//! (revision 28091c1532e869bc7aee91fc0aef6b3e6fd0b2e0): a Qwen3.5 hybrid VLM
+//! (`vlm_config`) plus a flow-matching planning expert (`expert_config`), with
+//! trajectory and sampler constants at the top level. Parsing here keeps the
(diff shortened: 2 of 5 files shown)


qhy991:
Closing in favour of #78, which consolidates the three device lines onto current `main`.

Carried over from here: `doc/qwen-drive-performance.md` with its measured 4090 numbers, which #78's device map cites as that board's row. The model-code items were already folded into #72's line before this consolidation started — #72's own description records which it took and which it measured and declined.

Measured, so it does not get lost: the seven identifiers this branch has that the consolidated line does not — `DecodeWorkspace`, `PrefillGdnScratch`, `concat_rows`, `linear_checkpoint_into`, `prefill_gdn_scratch`, `rope_row`, `views` — are the explicit persistent-workspace approach, which #72 replaced with per-call allocation behind `APXINF_CUDA_ALLOC_CACHE` after measuring the two on Orin. That measurement was on Orin, not on a 4090, so if the allocation cache behaves differently there this is the thing to look at first.

The 4090 has not been re-run since the consolidation; see the note on #51.

PR #74, opened 2026-09-16
docs(jetson): measured roofline for sm_87 and sm_101, and a tuning-store trap costing 12.6%

qhy991:
A measured roofline for two Jetson parts, the benchmark that produced it, and two findings that came out of taking the measurement seriously. Device-scoped rather than model-scoped: the tuning-store finding applies to every model in the tree, not just Qwen-Drive.

Separate from #72 (Qwen-Drive on Orin) and #73 (Qwen-Drive on RTX 4090) — this one is about the hardware and the tuning infrastructure.

## Why measure instead of quoting

Orin's commonly quoted 204.8 GB/s is the LPDDR5 pin rate. A kernel gets **152.8 GB/s**. That gap is the difference between reading Qwen-Drive's decode as "74% of the roof, keep optimising" and "at the roof, stop": decode moves 8.412 GB of weights per token, which at the measured bandwidth is a 55.1 ms floor and measures 55.2.

`scripts/bench_device_roofline.cu` measures read and copy bandwidth, BF16 tensor GEMM, and FP32 FMA against instruction-level parallelism. One file, builds with nvcc and cuBLAS, prints the numbers `doc/jetson-roofline.md` is built from.

| | Orin sm_87 | Thor sm_101 |
|---|---:|---:|
| SMs, clock | 16 @ 1300 MHz | 14 @ 1530 MHz |
| L2 | 4 MB | 24 MB |
| read (2GB, past L2) | **152.8 GB/s** | **257.0 GB/s** |
| BF16 tensor GEMM | **33.4–34.1 TFLOPS** | **135.2 TFLOPS** |
| FP32 FMA (flat in ILP) | 4.8–5.2 TFLOPS | 5.1–5.3 TFLOPS |

## A correction, recorded rather than quietly dropped

An earlier reading of this data reported a 4.5x FP32 dependency cliff — 1.18 TFLOPS at four independent chains rising to 5.15 at thirty-two — and used it to explain why the GDN scan kernels are slow, and to claim a 4.44x Thor advantage on exactly that pattern.

**There is no cliff.** A Jetson starts a cold process at a low DVFS point and takes about a second to reach its clock ceiling, which covered the first two points of the sweep. Warmed, the rate is flat:

| independent FMA chains per thread | 4 | 8 | 16 | 32 |
|---|---:|---:|---:|---:|
| cold, from process start | 1.18 | 1.14 | 4.07 | 5.15 |
| after a warm-up | 4.98 | 4.82 | 5.06 | 5.19 |

The cold row is clean enough to look like a real mechanism and support a confident wrong explanation. The benchmark now warms the device before every timed section and the header says why. Both the document and the benchmark carry the correction so the next person does not rediscover it the same way.

Also on Tegra: `cudaDevAttrMemoryClockRate` does not report the LPDDR rate. Computing a theoretical bandwidth from it gives 83.2 GB/s on a part that measures 152.8.

## Tuning stores are silently per-CUDA-version, and it costs 12.6%

`TuningPaths::for_cuda` keys the store by hardware alone — `configs/tuning/nvidia/<hardware>/tactics.json` — while the file header carries a single cuBLAS version. On a device whose toolkit does not match, **every record is rejected** and the model runs on cuBLAS defaults behind one warning line.

Measured on one board, one binary, four scenes, 64 tokens:

| store | s/scene |
|---|---:|
| committed store (recorded on CUDA 12.6), rejected on 13.2 | 7.12 |
| regenerated by one autotune pass on CUDA 13.2 | **6.22** |

The same tactics were worth 2.4% on CUDA 12.6. The 12.6% here is not the tuning being more valuable — it is cuBLAS 13's default heuristic being much worse for these shapes than 12.6's. Every JetPack R39 user of the committed store is paying it and sees only `ignored 65 incompatible CUDA tuning record(s)`.

Regenerating takes one inference with `APXINF_QWEN_AUTOTUNE=1`. Carrying both versions in the repository would mean putting the cuBLAS version in the store path, which relocates every existing store (rtx4090-sm89, thor-sm110, thor-sm101, orin-sm87); that is a decision for whoever owns the tuning layout, not something taken here.

## Where Qwen-Drive sits on Orin, for context

| stage | time | achieved | of its roof |
|---|---:|---:|---|
| decode | 55.2 ms/token | 152.4 GB/s | **99.7% of read** |
| prefill text GEMMs | 642 ms | 32.4 TFLOPS | **95–97% of BF16** |
| GDN scan (three kernels) | 874.8 ms | 0.50 TFLOPS / 17.7 GB/s | 10% / 12% |

Two of the three parts are finished against the hardware. The scan is bound by neither roof, so latency, occupancy or instruction mix — and with the cliff explanation withdrawn, **the mechanism is not established**. The document says so rather than offering a replacement guess.

## Incidental

The same branch on CUDA 12.6 and on 13.2 — different cuBLAS major version, store rejected on one, therefore different kernels selected — produces identical gate results: VQA first differs from the reference at index 121 with token 357, direct at 0.08056783676147461, reasoning in its known 0.0403/0.0806 pair. That is the strongest evidence so far that Qwen-Drive's divergence from the reference is BF16 rounding amplified through the GDN recurrence rather than a defect in one kernel or an artifact of one platform.


qhy991:
Closing in favour of #78, which carries this commit cherry-picked rather than restated, so `doc/jetson-roofline.md` and `scripts/bench_device_roofline.cu` arrive unchanged.

A closing section reads it against a third board, sm_110, and closes two things this left open:

- **The Orin bandwidth number reproduces.** 152.8 GB/s here; 154.4 GB/s measured independently on orin2 with a different program. And the `cudaDevAttrMemoryClockRate` warning was hit again — the same attribute produced 83.2 GB/s for Orin on a second probe — so it is repeated where the next person will see it.
- **The GDN scan's mechanism.** This says "the mechanism is not yet established" and warns that an earlier ILP explanation was a DVFS artifact. It is instruction mix, and specifically the unit: all four of the scan's inner products are GEMM-shaped and were running as scalar fp32 on the CUDA cores. On sm_110, moving `gdn_chunk_state` to tensor cores takes it from 452.3 ms to 187.8 ms per scene, and it is safe to move because the right-hand operand of every one of those products is already on the BF16 grid, so only the left needed splitting into two BF16 terms — 1.422947e-3 against the scalar form's 1.418808e-3 on an fp64 reference, where rounding rather than splitting costs 3.5x.

The tuning-store trap gets its other half too. The 12.6% here is what regenerating recovers on Orin; on sm_110 the same trap appears — 116 records rejected, recorded under CUDA 13.0 on a 13.2 board — but a full regeneration is worth 2.5% and all of it sits in the six decode records, which change the batch-1 GEMV tactic and drop VQA token agreement 27%. The seven prefill records are accuracy-neutral and worth 0.05%; only those ship. So the warning stands, and "regenerate the store" is not unconditionally free — check the result with an operator oracle or `precision_probe.py`, because the four-mode gate will not show you what it cost.

PR #77, opened 2026-09-17
Add attention kernel to cuda-new

ChenxiHe231:


crates/apxinf-cuda-new/native/adapters/attention/autotune.cu
zealoct, line 1: files that do not use CUDA syntax should not be suffixed with .cu

crates/apxinf-cuda-new/native/framework/autotune.h
zealoct, line 1: add tests for framework part of the backend

crates/apxinf-cuda-new/native/adapters/attention/autotune.cpp
ChenxiHe231: Addressed in 974e7aa. Translation units that neither contain nor transitively include CUDA syntax now use .cpp, and build.rs plus the Attention/GEMM build fingerprints were updated accordingly. Files that include kernel definitions or launch syntax remain .cu. Verified with a full CUDA build on Thor SM110.

crates/apxinf-cuda-new/native/framework/autotune.h
ChenxiHe231: Addressed in 974e7aa. Added independent black-box contracts for Registry exact identity matching, recipe DB exact/wrong/update/corrupt behavior, autotune filtering and failure handling, real CUDA timing in forward/reverse configuration order, and graph-safe winner capture. On Thor SM110 the 4 new tests pass; the full library suite is 52 passed / 0 failed / 1 pre-existing ignored, and the three integration targets are 13 passed.

PR #78, opened 2026-09-17
feat(qwen-drive): Qwen-Drive on RTX 4090, Jetson AGX Orin and Jetson AGX Thor

qhy991:
## Current review-round-2 status — `5d7d785` (2026-09-20)

Implementation corrections and additional evidence are submitted for review. **The historical first-divergent-operator localization requirement is still open; this is not a merge-approval claim.** See the [round-2 review response](https://github.com/infinigence/ApxInf/pull/78#issuecomment-5749250342).

- **New code fixes (`ed77678`):** reject zero `layers_per_kv` before division/remainder; bind projection layout once and remove the weights-to-executor dependency; load only the VLM through the text registry, with planner loading explicit.
- **Current implementation checks:** 12 Qwen model unit tests and a real CUDA/checkpoint registry-isolation test pass. The latter proves that a corrupt adjacent planner cannot break text loading, while explicit planner loading reports its error. Module-boundary checks pass.
- **Current all-mode fill checks:** zero, dirty and poison, three supported modes and four scenes: **32/32 pairwise complete-output comparisons are byte-identical**, and all 16 outputs match the previous fixed head. The clean rebuilt `7a630a3` parent is only **6/32**, reproducing the historical sensitivity rather than relying on an archived library.
- **Cached LayerNorm regression (`5d7d785`):** maintained 512/1024/2048-column coverage passes CPU references and Compute Sanitizer with 0 errors/0 warnings. An identical private probe against parent headers exposes 9 race-error groups; fixed headers expose none.
- **Remaining causal boundary:** the first real vision LayerNorm has identical complete inputs/outputs across the captured cases. A matched 64-token, same-script control also matches the captured parent-zero prefix; differences from the longer full probe cannot be attributed to instrumentation alone. The first divergent operator in the failing full-model runs is still not located. Neither this race nor the FA2 symbol collision is claimed to explain every historical discrepancy.
- **Review accounting:** 11/12 inline threads remain resolved, but general PR comments also contain review requirements; the count is not evidence that the remaining merge blocker is closed. Existing model-name aliases remain a maintainer decision. No new performance claim or current-head Thor/Orin full-model qualification is made.

## Previous fixed-head validation — `d12390a` (historical)

The following checks were recorded at d12390a, which integrated main a1a37af. They remain historical evidence; the current status and remaining review gate are stated above. See the [previous fixes and validation summary](https://github.com/infinigence/ApxInf/pull/78#issuecomment-5749047008).

- **Review fixes:** remaining diagnostic GPU readbacks/digest work are guarded; both planning paths receive the configured `num_steps`. Four new default/override policy tests pass. See the [readback reply](https://github.com/infinigence/ApxInf/pull/78#discussion_r4056644729) and [planning reply](https://github.com/infinigence/ApxInf/pull/78#discussion_r4056644789).
- **RTX 4090:** 143 CUDA tests and 5 architecture tests pass, with no failures or ignored tests. The legacy FP8 TN layout is fixed, including F16/BF16 output and changed-weight plan reuse checks. Unsupported dynamic rowwise FP8 is explicitly tested as unsupported, not claimed to be implemented.
- **FA2 stability:** dedicated reference-math kernels now have distinct entrypoints from generic fast-math kernels. Compiled weak-stub collisions are reduced from 10 to 0. Across all three supported modes and four scenes, 16 complete token/trajectory arrays are byte-identical with output clearing on versus off. Cache-on, cache-off, EAGER loading and early native import also produce identical complete 32-token controls, without `CUBLAS_WORKSPACE_CONFIG`.
- **Latest main compatibility:** PR-specific cached LayerNorm and softmax follow #80's scratch-reuse barriers; max reduction handles negative inputs. The standalone reduction/softmax/SDPA regression passes Compute Sanitizer racecheck on both 4090 and Thor (0 errors, 0 warnings). Model-family and PI0.5 boundary checks pass.
- **Inline review state then:** 12 threads total, 11 resolved. This did not account for the additional findings in general PR comments, now tracked in the current status above.
- **Performance observation:** three alternating rounds against the archived d9b700c native library give geomean **0.99161468×** (about 0.85% slower), including one 0.90284× scene-round. All samples are retained; this is not a speedup claim or a pass of the original 5% optimization gate.
- **Limits:** full HF numerical parity and perception remain outside this qualification. Current-head Thor evidence is compilation plus the focused reduction regression, not a new full-model run; Orin was not rerun on this head. Older device results below must not be read as validation of this revision.

The sections below retain historical development measurements; they are not the current post-merge qualification status.

---

Qwen-Drive-1.0-4B across **RTX 4090 (sm_89)**, **Jetson AGX Orin (sm_87)** and **Jetson AGX Thor (sm_110)**, consolidated into one branch.

Supersedes #51, #72, #73 and #74, whose content is carried here in full, which are four snapshots of one implementation rather than four implementations: no two are ancestors, but Orin's line differs from the 4090's by 5 files in the model code and 20 in the CUDA crate, and every file that looked exclusive to one side turned out to be merge-base staleness — both 4090 branches were cut before #54 moved robot and environment adaptation out of the engine. Merged onto current `main`, so this is independently mergeable.

#74's roofline work is cherry-picked rather than restated, and a new closing section reads it against sm_110: it reproduces its Orin bandwidth number to within 1%, answers the question it left open about the GDN scan's mechanism, and gives its tuning-store trap the other half of its story.

`doc/qwen-drive-devices.md` is the map: where each board stands, which of the three layers a per-device decision belongs in, and what has and has not been re-measured.

## Result

Four public WOD_E2E VQA scenes, batch 1, BF16, 64 generated tokens, one warmup and three measured requests per scene, warm median.

| | sm | per scene | against its own baseline |
|---|---|---:|---|
| RTX 4090 | sm_89 | 3.580 s | 1.02x (#73's pilot) |
| Jetson AGX Orin | sm_87 | 6.014 s | 1.09x, from 6.609 s (#72) |
| **Jetson AGX Thor** | **sm_110** | **3.505 s** | **1.4708x**, from 5.196 s |

Thor is the board this consolidation re-measured, and it is where the new work is. Its fixed cost goes 2.498 → 1.083 s, 57% off; decode moves 8.9% and should not move much further — it reads 8.41 GB of weights per token at about 250 GB/s against this board's *measured* 259.7 GB/s, so it is at 96% of the wall.

## Two things about Thor's starting point

Both are reasons this branch exists.

**The previous head does not build there.** `mha_f16`'s CUTLASS FMHA branch is compiled for the first time on an SM100-family device — every earlier target is outside it, so that code had never been through a compiler — and it does not survive: three undefined identifiers and an arity mismatch in `attention.rs`.

**With that fixed it still does not run.** The chunk-state block was set to 1024 threads unconditionally from an Orin measurement; on Thor that kernel's register demand at 1024 exceeds the budget and the launch returns `CUDA 701`. The policy table below carries the width per family and falls back rather than launching something invalid.

## Where per-device decisions live now

Three layers, and a change belongs to exactly one.

**Compile time — which kernels exist.** The family predicates in `build.rs` over `build_support/cuda_arch.rs`. Adding a board that shares a family is one arm of one predicate; that is how #60 added GB10 and how the head-64 and head-256 FA2 dispatch adapters reach Thor here.

**Run time — which path runs, with what constants.** Seven per-device decisions had accumulated inside `custom_kernels.cu`, each with its own `getenv` lambda and its own `>= 10 ? a : b`, on top of a `device_capability_major()` that re-queried the driver for a fact `CudaDeviceCaps` already holds and already classifies. They now come from one `#[repr(C)]` table in `kernels/gdn_policy.rs`, keyed by `CudaArchFamily`, with a per-field environment override for re-sweeping a new board without a rebuild. The adapter reads it and decides nothing.

That is where they belong: `CudaDeviceCaps` says in its own doc comment that it carries hardware facts only, and none of these is one — every one was swept on a board and came out different on the next.

| | sm80 family | sm100 family |
|---|---:|---:|
| chunk-state tile | 8 | 4 |
| chunk-gemm tile | 32 | 4 |
| recurrence split | 1 | 4 |
| chunk-state scan | scalar fp32 | two BF16 terms on tensor cores |

The chunk-state curve does not merely shift: a tile of 16 is 8% off the optimum on Orin and 26% off on Thor.

**Data — which tactic each shape uses.** `configs/tuning/<vendor>/<family>-sm<N>/`. Four stores ship and none needed unifying; that layer was already right.

## What is new for Thor

| change | per scene | why it was there to take |
|---|---|---|
| build fix for `mha_f16`'s FMHA branch | builds at all | never compiled on any prior device |
| head-256 FA2 on the SM100 family | −8.8% | the gate was on the dispatch adapters, not the kernels, which were already compiled |
| GDN tiles and splits from the device | −3.7% | Orin's sweeps are not Thor's |
| vision head-64 attention onto FA2 | −13.9% | the fallback was 628.6 ms of fp32 on CUDA cores |
| **GDN chunk-state scan on tensor cores** | **−7.1%** | 452 ms of GEMM-shaped work in scalar fp32 |

The last is the largest single change and the one that needed the most care. All four of the scan's inner products have a right-hand operand already on the BF16 grid, so rounding the left one to BF16 looked free — it is not, costing 3.5x against an fp64 reference. Carrying the left operand as two BF16 terms makes every partial product exact and lands at 1.422947e-3 against the scalar form's 1.418808e-3 for the same speed.

## Accuracy

The four-mode gate cannot decide these questions. It reports scene 0's maximum trajectory error, which on this checkpoint only ever takes the two adjacent BF16 output ULPs 0.0403 and 0.0806, so it flips on changes that move nothing — and it did, twice, in opposite directions. Two instruments were added:

- `control/precision_probe.py`: all four scenes of all three runnable modes, reporting token agreement and error distributions rather than one maximum.
- **fp64 operator oracles** in `tests/operators.rs` for the four operators this branch touches — vision segmented MHA, the GDN decode recurrence, the chunk-state scan, the chunk GEMM — each comparing the operator on one fixed input against a double-precision reference of its own maths.

The oracles decided every contested case. They overturned the gate on the vision attention route (FA2 is the *most* accurate of three at 2.305103e-3, not the least), chose between two tensor-core forms the end-to-end numbers could not separate, and kept two working optimisations switched off.

End to end on Thor, against the scalar baseline: VQA token agreement 0.325 → 0.609, direct trajectory mean/rms/p99 0.013483/0.036036/0.161133 → 0.011252/0.027358/0.097656. The gate's declared state is unchanged; `perception` remains the declared pending gap.

## Built, measured, left off

Each behind a switch that defaults off, with the number that rejected it in its commit message: autotuned decode GEMM tactics (2.5%, drops VQA agreement 27%), the tensor-core chunk GEMM (1.9%, `APXINF_GDN_CHUNK_GEMM_WMMA=3`), the tensor-core raw-attention term (0.7%, `APXINF_GDN_ATTN_RAW_WMMA=1`).

A fourth, a bit-exact packed weight format at 13.0625 bits against 16, is no longer in the diff at all. Its GEMV wins above n=36864 and loses 2.7x below it, which is where decode lives, so on review it was moved to `devlocal/qwen-drive/` rather than shipped switched off.

## On a megakernel or persistent runtime

Both sides are measured. Decode is 2428 ms over 35181 launches, of which four cuBLAS GEMV families are 2189 ms — 90.2% — leaving 159 ms a fusion could absorb, 4.1% of the scene. To fuse into those GEMVs they have to be ours, and `APXINF_PLAIN_GEMV=1` measures what ours is worth in the engine: 109.362 ms/token against cuBLAS's 38.100. Being 5% off on 2189 ms already costs 109 ms and cancels two thirds of the prize.

## Also here

`build(cuda): make the nvcc step incremental` — the build script re-ran nvcc for all 23 CUDA translation units on every rerun, about 40 minutes on Thor, for editing one line of Rust. It now keeps a dependency list and an argv stamp beside each object. First build unaffected, subsequent ones 80 seconds. Every measurement in this branch depended on it.

`doc/qwen-drive-devices.md` is the one document this PR adds: where each kind of per-device decision belongs, what the constants are, and which instrument decides a correctness question. The measurement evidence behind every number quoted here -- rooflines, kernel tables, per-device sweeps, and the standalone probe programs -- is under `devlocal/qwen-drive/`, out of the review diff per `AGENTS.md`.

## Earlier device validation (before the latest main merge)

All three boards have now been run, same scenes, same script, same two environment flags. `doc/qwen-drive-devices.md` has the full numbers.

**Correctness reproduces.** Orin returns what #72 and #74 recorded digit for digit — VQA first differing at index 121 with token 357, direct at 0.08056783676147461. The 4090 returns *the same seventeen digits* for direct on different silicon and a different toolkit, and Thor — the one board whose policy row puts the chunk-state scan on tensor cores — differs only in the last of them. That is the shape the divergence should have if it is BF16 rounding amplified through the GDN recurrence rather than a defect in one kernel.

| | VQA first difference | direct |
|---|---|---|
| Orin sm_87 | scene 0, index 121, token 357 | 0.08056783676147461 |
| RTX 4090 sm_89 | scenes 0 and 1 pass; scene 2 at index 487 | 0.08056783676147461 |
| Thor sm_110 | scene 0, index 106, token 5459 | 0.08056640625 |

**Thor: 1.4708x**, four-mode gate in its declared state, `precision_probe.py` unchanged, all four fp64 oracles returning the same numbers as before the merge, `gdn_policy` unit tests passing.

**RTX 4090: 1.78x faster**, not unchanged as expected — 2.5831 / 2.6048 / 2.5942 / 2.5676 s per scene becomes 1.4513 / 1.4536 / 1.4531 / 1.4536, alternating rounds on one machine. #73 is that board's own older line, with neither #72's work nor the twelve commits `main` has taken since, and consolidating hands it both. Accuracy improves too: direct trajectory p99 halves, 0.322266 → 0.161133.

**Orin: the 5.4% regression was a dirty baseline, and is resolved.** Profiling both wheels put 316 of the 340 ms on four kernels, and the launch shapes named the cause: `gdn_chunk_state` ran at block 512 with 64 registers a thread on one side and block 1024 with 48 on the other, because the baseline's source carried `__launch_bounds__` that this branch did not. Those launch bounds were **in no commit anywhere** — the baseline wheel had been built from a working tree on that board with 789 uncommitted lines in it, and #72's own head `80b0ecc` does not have them either. Nothing in the repository was slower than it had been.

That working tree also turned out to hold a real ~5% of Orin optimisation living on one board, so it is now folded in: the three GDN chunk kernels tile over columns instead of rows, they carry launch bounds as a register cap, the attention term accumulates a 4x4 register tile, and the residual add folds into the norm that consumes it. Orin is **6.07% faster** than before the fold and 3.12% ahead of the tree it recovers; Thor gains 0.94%, less because its policy row puts the chunk-state scan on tensor cores so the largest of the four pieces never runs there. The operator oracles return the same digits they did before the fold and the precision probe is byte-identical, first differing VQA index `[121, 88, 445, 32]` included.

The tuning-store caveat is fixed rather than documented. A tactic is only valid for the libraries that measured it, and **all three boards** were loading a store recorded on a different toolkit, rejecting it record by record, and silently falling back to the untuned heuristic — Orin 12.6 against 13.2, Thor 13.0 against 13.2, the 4090 12.8 against 12.3. `TuningPaths::resolve_for_cuda` now prefers `<hardware>/cuda<major>.<minor>-cublas<major>.<minor>/tactics.json`, keyed by the same truncation the loader compares, and falls back to the unqualified file only when its header matches. Nothing relocates.

## Review round

The original review round contained 10 threads (11 initial comments counting one follow-up), addressed by the responses and changes in `5345d56`. Nine original threads are resolved; the pre-existing `qwen3_vl` / `qwen3vl` alias question remains open. Everything not on a supported path left the diff for `devlocal/`: the probes and the benchmark script, seven of the eight new docs, the packed-weight GEMV, and the perception scaffold with its cuDNN convolution, pooling and group-norm operators — the sweep also found `nchw_to_nhwc_bf16`, `nhwc_to_nchw_bf16`, `batch_relu_bf16`, `global_mean` and `partial_rope_table` had no callers at all. `merge_rows` was a kernel that copied every element to do what `Tensor::reshape` does for free, so the call site uses reshape. The PR is now 60 files and +27706 lines, from 87 and +32317.

Two of the comments were bugs. The development diagnostics ran on every request and several of the lines carried prompt and generated token ids; all 27 now go through one `qdiag!` macro whose body is the `APXINF_QWEN_DIAG` check, and a normal request writes nothing to stderr from this model on either Jetson. The allocation cache could hand a block to a second stream while the first was still reading it — the free list was process-wide and keyed by `(device, size)` alone — and is now keyed by `(device, stream, size)`, with blocks that have no stream never cached; measured at 6.55 s/scene with the cache and 17.5 s without, so the stream keying costs none of what it was built for.

Taking the comments is free on both boards that could be measured: geomean 1.0016 on Orin and 1.0030 on Thor, three alternating rounds each.

**Known gap:** two results are "cause unidentified" because `ncu` cannot run on the Thor board — `ERR_NVGPUCTRPERM`, one root-level driver setting away.

### DIFF
--- a/crates/apxinf-model/src/qwen_drive/mod.rs
+++ b/crates/apxinf-model/src/qwen_drive/mod.rs
@@ -1,0 +1,24 @@
+//! Qwen-Drive-1.0 model family: checkpoint configuration, weight mapping,
+//! and the planning-expert executor.
+//!
+//! Model family isolation: every Qwen-Drive-specific type lives behind this
+//! module; nothing here is mixed into the generic llama / qwen3vl paths.
+//!
+//! Device status (honest accounting):
+//!
+//! * `planner` is an explicitly host-side f32 correctness scaffold (bf16
+//!   rounding at the reference's semantic rounding sites), not native GPU
+//!   deployment; its device exit criterion is the CUDA executor gate from the
+//!   execution ledger.
+//! * The VLM scene-cache producer is pending reference qualification, so
+//!   end-to-end planning activates only when both halves land.
+
+pub mod config;
+pub mod planner;
+pub mod weights;
+
+pub use config::{
+    PlanningExpertConfig, QwenDriveConfig, QwenDriveTextConfig, QwenDriveVisionConfig,
+};
+pub use planner::{ExpertConditioning, PlanningExpertModel, SceneCache};
+pub use weights::{QwenDriveExpertWeights, QwenDriveVlmWeights};
--- a/crates/apxinf-model/src/lib.rs
+++ b/crates/apxinf-model/src/lib.rs
@@ -11,6 +11,7 @@ pub mod auto;
 pub mod profiling;
 pub mod pi05;
 pub mod qwen3vl;
+pub mod qwen_drive;
 pub mod vla;
 mod walloss;
 
@@ -27,6 +28,7 @@ pub use auto::{AutoModel, LoadOptions, LoadedModel, ModelPrecision, SyntheticWei
 pub use profiling::GenerationProfile;
 pub use pi05::{Pi05Config, Pi05PerformanceProfile};
 pub use qwen3vl::{GeneralQwen3VL, Qwen3VLConfig, Qwen3VLTextWeights};
+pub use qwen_drive::QwenDriveConfig;
 pub use vla::{
     Action, ImageLayout, InferenceSpec, InitialLatent, Observation,
     PreparedInference, VisionObservation, VlaContract, VlaRequest, VlaRuntime,
--- a/crates/apxinf-model/src/qwen_drive/config.rs
+++ b/crates/apxinf-model/src/qwen_drive/config.rs
@@ -1,0 +1,553 @@
+//! Qwen-Drive-1.0 configuration, parsed from the checkpoint `config.json`.
+//!
+//! The schema mirrors `configuration_qwen_drive.py` in the pinned reference
+//! (revision 28091c1532e869bc7aee91fc0aef6b3e6fd0b2e0): a Qwen3.5 hybrid VLM
+//! (`vlm_config`) plus a flow-matching planning expert (`expert_config`), with
+//! trajectory and sampler constants at the top level. Parsing here keeps the
+//! loader crate model-agnostic and matches "each model owns its config".
+
+use std::path::Path;
+
+use apxinf_core::{Error, Result};
+
+/// Text-decoder geometry of the Qwen3.5 VLM. The decoder is hybrid: layers
+/// listed as `full_attention` carry a standard GQA KV cache (these are the
+/// caches the planning expert reads), while `linear_attention` layers are
+/// gated delta net (GDN) linear-attention layers with a recurrent state.
+#[derive(Clone, Debug)]
+pub struct QwenDriveTextConfig {
+    pub hidden_size: usize,
+    pub intermediate_size: usize,
+    pub n_layers: usize,
+    pub n_heads: usize,
+    pub n_kv_heads: usize,
+    pub head_dim: usize,
+    pub vocab_size: usize,
+    pub max_position_embeddings: usize,
+    pub rms_norm_eps: f32,
+    /// Qwen3.5 emits a per-head sigmoid gate on the attention output.
+    pub attn_output_gate: bool,
+    /// Per-layer type names: `full_attention` or `linear_attention`.
+    pub layer_types: Vec<String>,
+    // GDN geometry (only meaningful for `linear_attention` layers).
+    pub linear_num_key_heads: usize,
+    pub linear_key_head_dim: usize,
+    pub linear_num_value_heads: usize,
+    pub linear_value_head_dim: usize,
+    pub linear_conv_kernel_dim: usize,
+    // Partial interleaved mRoPE.
+    pub rope_theta: f32,
+    pub partial_rotary_factor: f32,
+    pub mrope_section: [usize; 3],
+    pub mrope_interleaved: bool,
+    pub tie_word_embeddings: bool,
+    pub eos_token_id: u32,
+}
+
+impl QwenDriveTextConfig {
+    /// Rotary channel count of the partial RoPE (64 of 256 for Qwen3.5-4B).
+    pub fn rotary_dim(&self) -> usize {
+        (self.head_dim as f32 * self.partial_rotary_factor).round() as usize
+    }
+
+    /// Indices of the full-attention layers, in layer order. These are exactly
+    /// the caches the planning expert consumes (config.full_attention_layers).
+    pub fn full_attention_layers(&self) -> Vec<usize> {
+        self.layer_types
+            .iter()
+            .enumerate()
+            .filter(|(_, kind)| kind.as_str() == "full_attention")
+            .map(|(index, _)| index)
+            .collect()
+    }
+
+    pub fn is_full_attention(&self, layer_index: usize) -> bool {
+        self.layer_types
+            .get(layer_index)
+            .map(|kind| kind.as_str() == "full_attention")
+            .unwrap_or(false)
+    }
+
+    /// Validate cross-field invariants the executor relies on.
+    pub fn validate(&self) -> Result<()> {
+        if self.layer_types.len() != self.n_layers {
+            return Err(Error::Other(format!(
+                "qwen_drive text config: layer_types has {} entries but num_hidden_layers is {}",
+                self.layer_types.len(),
+                self.n_layers
+            )));
+        }
+        for (index, kind) in self.layer_types.iter().enumerate() {
+            if kind != "full_attention" && kind != "linear_attention" {
+                return Err(Error::Other(format!(
+                    "qwen_drive text config: unknown layer type {kind:?} at index {index}"
+                )));
+            }
+        }
+        let rotary = self.rotary_dim();
+        if rotary == 0 || rotary % 2 != 0 || rotary > self.head_dim {
+            return Err(Error::Other(format!(
+                "qwen_drive text config: invalid rotary dim {rotary} for head dim {}",
+                self.head_dim
+            )));
+        }
+        let pairs: usize = self.mrope_section.iter().sum();
+        if pairs != rotary / 2 {
+            return Err(Error::Other(format!(
+                "qwen_drive text config: mrope_section {:?} must sum to {} frequency pairs",
+                self.mrope_section,
+                rotary / 2
+            )));
+        }
+        Ok(())
+    }
+}
+
+/// Vision tower geometry (ViT, 24 blocks, hidden 1024, merger to 2560).
+#[derive(Clone, Debug)]
+pub struct QwenDriveVisionConfig {
+    pub depth: usize,
+    pub hidden_size: usize,
+    pub intermediate_size: usize,
+    pub num_heads: usize,
+    pub head_dim: usize,
+    pub patch_size: usize,
+    pub temporal_patch_size: usize,
+    pub in_channels: usize,
+    pub spatial_merge_size: usize,
+    pub num_position_embeddings: usize,
+    pub out_hidden_size: usize,
+}
+
+impl QwenDriveVisionConfig {
+    pub fn head_dim(&self) -> usize {
+        if self.head_dim != 0 {
+            self.head_dim
+        } else {
+            self.hidden_size / self.num_heads
+        }
+    }
+}
+
+/// Planning expert geometry (`PlanningExpertConfig` in the reference).
+#[derive(Clone, Debug)]
+pub struct PlanningExpertConfig {
+    pub hidden_size: usize,
+    pub intermediate_size: usize,
+    pub n_layers: usize,
+    pub n_heads: usize,
+    pub n_kv_heads: usize,
+    pub head_dim: usize,
+    /// Consecutive expert layers sharing one VLM scene cache.
+    pub layers_per_kv: usize,
+    pub rms_norm_eps: f32,
+    pub time_embed_dim: usize,
+    pub time_embed_scale: f32,
+    pub fourier_num_features: usize,
+    pub fourier_max_frequency: f32,
+    pub nav_command_classes: usize,
+    pub ego_status_dim: usize,
+    pub history_dynamics_dim: usize,
+    pub rope_theta: f32,
+    pub partial_rotary_factor: f32,
+    pub mrope_section: [usize; 3],
+}
+
+impl PlanningExpertConfig {
+    pub fn rotary_dim(&self) -> usize {
+        (self.head_dim as f32 * self.partial_rotary_factor).round() as usize
+    }
+
+    /// Number of VLM scene caches consumed (one per `layers_per_kv` layers).
+    pub fn num_kv_sources(&self) -> usize {
+        self.n_layers / self.layers_per_kv
+    }
+
+    pub fn validate(&self) -> Result<()> {
+        if self.n_layers % self.layers_per_kv != 0 {
+            return Err(Error::Other(format!(
+                "qwen_drive expert config: {} layers not divisible by layers_per_kv {}",
+                self.n_layers,
+                self.layers_per_kv
+            )));
+        }
+        let rotary = self.rotary_dim();
+        if rotary % 2 != 0 || rotary > self.head_dim {
+            return Err(Error::Other(format!(
+                "qwen_drive expert config: invalid rotary dim {rotary}"
+            )));
+        }
+        let pairs: usize = self.mrope_section.iter().sum();
+        if pairs != rotary / 2 {
+            return Err(Error::Other(format!(
+                "qwen_drive expert config: mrope_section {:?} must sum to {} frequency pairs",
+                self.mrope_section,
+                rotary / 2
+            )));
+        }
+        Ok(())
+    }
+}
+
+/// Full Qwen-Drive-1.0 configuration (`QwenDriveConfig` in the reference).
+#[derive(Clone, Debug)]
+pub struct QwenDriveConfig {
+    pub text: QwenDriveTextConfig,
+    pub vision: QwenDriveVisionConfig,
+    pub expert: PlanningExpertConfig,
+    pub image_token_id: u32,
+    pub video_token_id: u32,
+    pub vision_start_token_id: u32,
+    pub vision_end_token_id: u32,
+    // Trajectory geometry.
+    pub num_future_points: usize,
+    pub num_history_points: usize,
+    pub trajectory_point_dim: usize,
+    pub trajectory_hz: f32,
+    /// Per-channel normalization scale; the heading entry is the bfloat16
+    /// rounding of pi/2 (1.5703125), which is the value training saw.
+    pub trajectory_scale: [f32; 3],
+    // Flow-matching sampler.
+    pub num_inference_steps: usize,
+    pub noise_init_std: f32,
+    pub noise_seed: u64,
+    pub min_one_minus_t: f32,
+    // Image preprocessing budgets.
+    pub history_image_pixels: usize,
+    pub current_image_pixels: usize,
+    pub image_patch_size: usize,
+    pub image_temporal_patch_size: usize,
+    pub image_spatial_merge_size: usize,
+    // Reasoning-stage generation bounds.
+    pub max_reasoning_tokens: usize,
+    pub min_reasoning_tokens: usize,
+}
+
+impl QwenDriveConfig {
+    pub fn from_json_file(path: &Path) -> Result<Self> {
+        let raw = std::fs::read_to_string(path)
+            .map_err(|e| Error::Other(format!("read {}: {e}", path.display())))?;
+        Self::from_json_str(&raw)
+    }
+
+    pub fn from_json_str(s: &str) -> Result<Self> {
+        let v: serde_json::Value = serde_json::from_str(s)
+            .map_err(|e| Error::Other(format!("qwen_drive config json: {e}")))?;
+        let vlm = &v["vlm_config"];
+        if !vlm.is_object() {
+            return Err(Error::Other("qwen_drive config: missing vlm_config".into()));
+        }
+        let tc = &vlm["text_config"];
+        if !tc.is_object() {
+            return Err(Error::Other(
+                "qwen_drive config: missing vlm_config.text_config".into(),
+            ));
+        }
+        let layer_types: Vec<String> = tc["layer_types"]
+            .as_array()
+            .ok_or_else(|| Error::Other("qwen_drive config: missing text_config.layer_types".into()))?
+            .iter()
+            .filter_map(|entry| entry.as_str().map(str::to_owned))
+            .collect();
+        let rope = &tc["rope_parameters"];
+        let section = rope["mrope_section"]
+            .as_array()
+            .ok_or_else(|| {
+                Error::Other("qwen_drive config: missing rope_parameters.mrope_section".into())
+            })?;
+        if section.len() != 3 {
+            return Err(Error::Other(format!(
+                "qwen_drive config: mrope_section must have 3 entries, got {}",
+                section.len()
+            )));
+        }
+        let mrope_section = [
+            section[0].as_u64().unwrap_or(11) as usize,
+            section[1].as_u64().unwrap_or(11) as usize,
+            section[2].as_u64().unwrap_or(10) as usize,
+        ];
+        let text = QwenDriveTextConfig {
+            hidden_size: tc["hidden_size"].as_u64().unwrap_or(2560) as usize,
+            intermediate_size: tc["intermediate_size"].as_u64().unwrap_or(9216) as usize,
+            n_layers: tc["num_hidden_layers"].as_u64().unwrap_or(32) as usize,
+            n_heads: tc["num_attention_heads"].as_u64().unwrap_or(16) as usize,
+            n_kv_heads: tc["num_key_value_heads"].as_u64().unwrap_or(4) as usize,
+            head_dim: tc["head_dim"].as_u64().unwrap_or(256) as usize,
+            vocab_size: tc["vocab_size"].as_u64().unwrap_or(248320) as usize,
+            max_position_embeddings: tc["max_position_embeddings"].as_u64().unwrap_or(32768) as usize,
+            rms_norm_eps: tc["rms_norm_eps"].as_f64().unwrap_or(1e-6) as f32,
+            attn_output_gate: tc["attn_output_gate"].as_bool().unwrap_or(true),
+            layer_types,
+            linear_num_key_heads: tc["linear_num_key_heads"].as_u64().unwrap_or(16) as usize,
+            linear_key_head_dim: tc["linear_key_head_dim"].as_u64().unwrap_or(128) as usize,
+            linear_num_value_heads: tc["linear_num_value_heads"].as_u64().unwrap_or(32) as usize,
+            linear_value_head_dim: tc["linear_value_head_dim"].as_u64().unwrap_or(128) as usize,
+            linear_conv_kernel_dim: tc["linear_conv_kernel_dim"].as_u64().unwrap_or(4) as usize,
+            rope_theta: rope["rope_theta"].as_f64().unwrap_or(10_000_000.0) as f32,
+            partial_rotary_factor: rope["partial_rotary_factor"].as_f64().unwrap_or(0.25) as f32,
+            mrope_section,
+            mrope_interleaved: rope["mrope_interleaved"].as_bool().unwrap_or(true),
+            tie_word_embeddings: tc["tie_word_embeddings"].as_bool().unwrap_or(true),
+            eos_token_id: tc["eos_token_id"].as_u64().unwrap_or(248044) as u32,
+        };
+        text.validate()?;
+
+        let vc = &vlm["vision_config"];
+        if !vc.is_object() {
+            return Err(Error::Other(
+                "qwen_drive config: missing vlm_config.vision_config".into(),
+            ));
+        }
+        let vision = QwenDriveVisionConfig {
+            depth: vc["depth"].as_u64().unwrap_or(24) as usize,
+            hidden_size: vc["hidden_size"].as_u64().unwrap_or(1024) as usize,
+            intermediate_size: vc["intermediate_size"].as_u64().unwrap_or(4096) as usize,
+            num_heads: vc["num_heads"].as_u64().unwrap_or(16) as usize,
+            head_dim: vc.get("head_dim").and_then(|x| x.as_u64()).map(|x| x as usize).unwrap_or(0),
+            patch_size: vc["patch_size"].as_u64().unwrap_or(16) as usize,
+            temporal_patch_size: vc["temporal_patch_size"].as_u64().unwrap_or(2) as usize,
+            in_channels: vc["in_channels"].as_u64().unwrap_or(3) as usize,
+            spatial_merge_size: vc["spatial_merge_size"].as_u64().unwrap_or(2) as usize,
+            num_position_embeddings: vc["num_position_embeddings"].as_u64().unwrap_or(2304) as usize,
+            out_hidden_size: vc["out_hidden_size"].as_u64().unwrap_or(2560) as usize,
+        };
+
+        let ec = &v["expert_config"];
+        if !ec.is_object() {
+            return Err(Error::Other("qwen_drive config: missing expert_config".into()));
+        }
+        let esection = ec["mrope_section"]
+            .as_array()
+            .ok_or_else(|| Error::Other("qwen_drive config: missing expert mrope_section".into()))?;
+        if esection.len() != 3 {
+            return Err(Error::Other(format!(
+                "qwen_drive config: expert mrope_section must have 3 entries, got {}",
+                esection.len()
+            )));
+        }
+        let expert = PlanningExpertConfig {
+            hidden_size: ec["hidden_size"].as_u64().unwrap_or(1024) as usize,
+            intermediate_size: ec["intermediate_size"].as_u64().unwrap_or(3584) as usize,
+            n_layers: ec["num_hidden_layers"].as_u64().unwrap_or(32) as usize,
+            n_heads: ec["num_attention_heads"].as_u64().unwrap_or(16) as usize,
+            n_kv_heads: ec["num_key_value_heads"].as_u64().unwrap_or(4) as usize,
+            head_dim: ec["head_dim"].as_u64().unwrap_or(256) as usize,
+            layers_per_kv: ec["layers_per_kv"].as_u64().unwrap_or(4) as usize,
+            rms_norm_eps: ec["rms_norm_eps"].as_f64().unwrap_or(1e-5) as f32,
+            time_embed_dim: ec["time_embed_dim"].as_u64().unwrap_or(128) as usize,
+            time_embed_scale: ec["time_embed_scale"].as_f64().unwrap_or(1000.0) as f32,
+            fourier_num_features: ec["fourier_num_features"].as_u64().unwrap_or(16) as usize,
+            fourier_max_frequency: ec["fourier_max_frequency"].as_f64().unwrap_or(16.0) as f32,
+            nav_command_classes: ec["nav_command_classes"].as_u64().unwrap_or(3) as usize,
+            ego_status_dim: ec["ego_status_dim"].as_u64().unwrap_or(8) as usize,
+            history_dynamics_dim: ec["history_dynamics_dim"].as_u64().unwrap_or(2) as usize,
+            rope_theta: ec["rope_theta"].as_f64().unwrap_or(10_000_000.0) as f32,
+            partial_rotary_factor: ec["partial_rotary_factor"].as_f64().unwrap_or(0.25) as f32,
+            mrope_section: [
+                esection[0].as_u64().unwrap_or(11) as usize,
+                esection[1].as_u64().unwrap_or(11) as usize,
+                esection[2].as_u64().unwrap_or(10) as usize,
+            ],
+        };
+        expert.validate()?;
+
+        let scale = v["trajectory_scale"]
+            .as_array()
+            .ok_or_else(|| Error::Other("qwen_drive config: missing trajectory_scale".into()))?;
+        if scale.len() != 3 {
+            return Err(Error::Other(format!(
+                "qwen_drive config: trajectory_scale must have 3 entries, got {}",
+                scale.len()
+            )));
+        }
+        let config = QwenDriveConfig {
+            text,
+            vision,
+            expert,
+            image_token_id: vlm["image_token_id"].as_u64().unwrap_or(248056) as u32,
+            video_token_id: vlm["video_token_id"].as_u64().unwrap_or(248057) as u32,
+            vision_start_token_id: vlm["vision_start_token_id"].as_u64().unwrap_or(248053) as u32,
+            vision_end_token_id: vlm["vision_end_token_id"].as_u64().unwrap_or(248054) as u32,
+            num_future_points: v["num_future_points"].as_u64().unwrap_or(50) as usize,
+            num_history_points: v["num_history_points"].as_u64().unwrap_or(16) as usize,
+            trajectory_point_dim: v["trajectory_point_dim"].as_u64().unwrap_or(3) as usize,
+            trajectory_hz: v["trajectory_hz"].as_f64().unwrap_or(10.0) as f32,
+            trajectory_scale: [
+                scale[0].as_f64().unwrap_or(165.0) as f32,
+                scale[1].as_f64().unwrap_or(25.0) as f32,
+                scale[2].as_f64().unwrap_or(1.5703125) as f32,
+            ],
+            num_inference_steps: v["num_inference_steps"].as_u64().unwrap_or(10) as usize,
+            noise_init_std: v["noise_init_std"].as_f64().unwrap_or(1.0) as f32,
+            noise_seed: v["noise_seed"].as_u64().unwrap_or(42),
+            min_one_minus_t: v["min_one_minus_t"].as_f64().unwrap_or(0.1) as f32,
+            history_image_pixels: v["history_image_pixels"].as_u64().unwrap_or(174080) as usize,
+            current_image_pixels: v["current_image_pixels"].as_u64().unwrap_or(921600) as usize,
+            image_patch_size: v["image_patch_size"].as_u64().unwrap_or(16) as usize,
+            image_temporal_patch_size: v["image_temporal_patch_size"].as_u64().unwrap_or(2) as usize,
+            image_spatial_merge_size: v["image_spatial_merge_size"].as_u64().unwrap_or(2) as usize,
+            max_reasoning_tokens: v["max_reasoning_tokens"].as_u64().unwrap_or(256) as usize,
+            min_reasoning_tokens: v["min_reasoning_tokens"].as_u64().unwrap_or(10) as usize,
+        };
+        config.validate()?;
+        Ok(config)
+    }
+
+    /// Cross-validate the VLM and the expert: the expert reads the VLM's
+    /// post-rotary K/V directly, so their attention geometry must match and
+    /// the number of exported caches must equal the expert's KV sources.
+    pub fn validate(&self) -> Result<()> {
+        let full = self.text.full_attention_layers();
+        if full.len() != self.expert.num_kv_sources() {
+            return Err(Error::Other(format!(
+                "qwen_drive config: {} VLM full-attention layers but the expert expects {} scene caches",
+                full.len(),
+                self.expert.num_kv_sources()
+            )));
+        }
+        if self.text.n_kv_heads != self.expert.n_kv_heads || self.text.head_dim != self.expert.head_dim {
+            return Err(Error::Other(format!(
+                "qwen_drive config: expert KV geometry ({} heads x dim {}) must match the VLM ({} x {})",
+                self.expert.n_kv_heads,
+                self.expert.head_dim,
+                self.text.n_kv_heads,
+                self.text.head_dim
+            )));
+        }
+        if self.num_history_points < 2 {
+            return Err(Error::Other(
+                "qwen_drive config: num_history_points must be at least 2".into(),
+            ));
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    // Abbreviated but schema-faithful excerpt of the released checkpoint
+    // config (experiment/qwen-drive-k3/checkpoint-configs/config.json).
+    const CHECKPOINT_CONFIG: &str = r#"{
+        "model_type": "qwen_drive",
+        "num_future_points": 50,
+        "num_history_points": 16,
+        "trajectory_point_dim": 3,
+        "trajectory_hz": 10.0,
+        "trajectory_scale": [165.0, 25.0, 1.5703125],
+        "num_inference_steps": 10,
+        "noise_init_std": 1.0,
+        "noise_seed": 42,
+        "min_one_minus_t": 0.1,
+        "max_reasoning_tokens": 256,
+        "history_image_pixels": 174080,
+        "current_image_pixels": 921600,
+        "image_patch_size": 16,
+        "image_temporal_patch_size": 2,
+        "image_spatial_merge_size": 2,
+        "expert_config": {
+            "hidden_size": 1024,
+            "intermediate_size": 3584,
+            "num_hidden_layers": 32,
+            "num_attention_heads": 16,
+            "num_key_value_heads": 4,
+            "head_dim": 256,
+            "layers_per_kv": 4,
+            "rms_norm_eps": 1e-05,
+            "time_embed_dim": 128,
+            "time_embed_scale": 1000.0,
+            "fourier_num_features": 16,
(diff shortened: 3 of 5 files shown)


probes/ablate.cu
```diff
@@ -0,0 +1,158 @@
+// Which part of the packed read costs the bandwidth? Four kernels with the
```
wong-1994: This whole "probes" directory shouldn't be commited. See agent.md.

doc/jetson-roofline.md
```diff
@@ -0,0 +1,247 @@
+# Measured roofline on Jetson, and what it says about Qwen-Drive
```
wong-1994: All these newly added .md files seem to be experiment report and should not be commited.
Only keep necessary instruction in /doc. Do not add new docs unless there's real difference between the usage of Qwen-drive and models (that is worth a head up or additional notes).

crates/apxinf-cuda/src/buffer.rs
```diff
@@ -40,10 +114,14 @@ unsafe impl Sync for CudaAllocation {}
...
 impl Drop for CudaAllocation {
     fn drop(&mut self) {
-        if !self.ptr.is_null() {
-            unsafe {
-                let _ = ffi::cudaFree(self.ptr);
-            }
+        if self.ptr.is_null() {
+            return;
+        }
+        if cache_put(self.ptr, self.len, self.device) {
```
ChenxiHe231: Cached allocation can be reused while still in flight

Dropping a buffer immediately publishes its pointer to a process-global cache keyed only by device and size. CUDA launches are asynchronous, so another context with a different stream can obtain this pointer before the original stream has finished using it. APXINF_CUDA_ALLOC_CACHE=1 can therefore alias live allocations during concurrent inference. Reuse needs stream ordering or an event-completion check.

crates/apxinf-model/src/qwen_drive/vision.rs
```diff
@@ -0,0 +1,405 @@
...
+    let pos_embeds = compute_pos_embeds(config, weights, ctx, grid_thw)?;
+    x = elementwise::add(ctx, &x, &pos_embeds)?;
+
+    // 2D rope position ids in the merge-block-major patch order.
+    let pos_ids = compute_vision_pos_ids(grid_thw, merge);
+    let pos_ids_dev = upload_u32(ctx, &pos_ids)?;
+
+    for (block_idx, block) in weights.blocks.iter().enumerate() {
+        // TEMP-DIAG (implement_r2, ungated in implement_r3): vision-block heartbeat every block; revert in the acceptance-bound revision.
+        eprintln!(
```
ChenxiHe231: This prints once per vision block on every request even when APXINF_QWEN_DIAG is unset. generate also has unconditional entry, prompt-tail, prefill and decode prints. These development probes should be removed or consistently guarded by diagnostics_enabled; otherwise normal inference floods stderr and exposes request token metadata.

probes/ablate.cu
ChenxiHe231: +1，Intermediate artifacts—such as experimental code, one-off scripts, debug probes, and benchmark results—must be placed in `devlocal/<feat-name>/`.

crates/apxinf-model/src/builtin.rs
```diff
@@ -17,6 +17,8 @@ pub fn register_builtin_models() {
     registry::register("llama", load_llama);
     registry::register("qwen3_vl", load_qwen3vl);
     registry::register("qwen3vl", load_qwen3vl);
+    registry::register("qwen_drive", load_qwen_drive);
```
wong-1994, line 20: Why do we need qwen_drive and qwen_drive_cuda?

crates/apxinf-model/src/builtin.rs
```diff
@@ -17,6 +17,8 @@ pub fn register_builtin_models() {
     registry::register("llama", load_llama);
     registry::register("qwen3_vl", load_qwen3vl);
```
wong-1994, line 18: Might not relevent to this pr but why do we have a qwen3_vl AND a qwen3vl?

crates/apxinf-model/src/qwen_drive/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+#[cfg(feature = "cuda")]
+pub mod backend;
+#[cfg(feature = "cuda")]
+pub mod device_weights;
+#[cfg(feature = "cuda")]
+pub mod expert;
+#[cfg(feature = "cuda")]
+pub mod general;
+#[cfg(feature = "cuda")]
+pub mod perception_depth;
```
ChenxiHe231: The public policy explicitly rejects perception in this revision, and this module also says that the BEV stack is not connected or verified. That makes these perception modules, together with the new cuDNN convolution, pooling, and normalization operators, unreachable from every supported request. Please split the perception scaffold and its operators into a follow-up PR when an end-to-end perception path exists. They are not required for VQA, direct planning, or reasoning planning.

crates/apxinf-cuda/src/kernels/gemm/packed.rs
```diff
@@ -0,0 +1,184 @@
...
+    base: CudaBuffer,
+    k: usize,
+    n: usize,
+}
+
+/// Keyed by the weight's device address and shape. Model weights live for the
+/// process, so an entry is valid for as long as it can be looked up.
+static CACHE: Mutex<Option<HashMap<(usize, usize, usize), PackedWeight>>> = Mutex::new(None);
+
+/// Whether the packed decode path is on. Off unless asked for.
```
ChenxiHe231: This subsystem is opt-in and disabled by default. It also retains an additional packed copy of the model weights, while the PR description says it loses substantially on the widths where most decode GEMVs live. It does not contribute to the default Qwen-Drive path or its claimed speedup. Please remove it from this PR and keep the experiment under devlocal/ or submit it separately once it is selected for production.

crates/apxinf-cuda/src/kernels/linear_attention.rs
```diff
@@ -0,0 +1,1263 @@
...
+        DType::BF16,
+        ctx.device_id(),
+        output,
+    ))
+}
+
+/// Partial rotary from precomputed BF16 cos/sin tables `[rows, rotary_dim]`.
+/// `x` is `[rows, heads, head_dim]`; leading `rotary_dim` channels rotate,
+/// the rest pass through.
+pub fn partial_rope_table(
```
ChenxiHe231: I could not find any production caller of partial_rope_table; only its Rust/FFI/adapter/kernel definitions reference it. Please remove this unused four-layer API, or add the missing production call and a test that demonstrates why it is needed in this PR.

crates/apxinf-cuda/src/kernels/linear_attention.rs
```diff
@@ -0,0 +1,1263 @@
...
+        input.shape().clone(),
+        DType::BF16,
+        ctx.device_id(),
+        output,
+    ))
+}
+
+/// Merge `factor` consecutive rows into one row:
+/// `[rows, cols] -> [rows / factor, cols * factor]`.
+pub fn merge_rows(ctx: &CudaContext, input: &Tensor, factor: usize) -> Result<Tensor> {
```
ChenxiHe231: The CUDA kernel preserves the exact contiguous element order and only changes [rows, cols] into [rows / factor, cols * factor]. The call site can use Tensor::reshape directly, without allocating another buffer, copying every element, adding an FFI entry, or launching a CUDA kernel. Please reuse the existing reshape operation.

probes/ablate.cu
qhy991: Done in `5345d56` — `probes/` is out of the diff and lives under `devlocal/qwen-drive/probes/`, which `.gitignore` already covers. `scripts/bench_device_roofline.cu` went with it for the same reason.

crates/apxinf-cuda/src/buffer.rs
qhy991: Good catch, and fixed in `5345d56`.

The free list was keyed by `(device, size)` alone and was process-wide, so exactly as you describe: a launch is asynchronous, and the pointer one context dropped could be taken by another context's stream with nothing ordering the two.

It is now keyed by `(device, stream, size)`. Within one stream the launches are ordered by definition, so reuse is safe; cross-stream reuse cannot happen. A buffer with no stream attached is never cached at all, which is why `CudaBuffer::alloc` no longer consults the cache and a new `alloc_on(ctx, …)` does — the hot path (`workspace::output_buffer`, the 43,811 malloc/free pairs this was built for) already has the context in hand.

Measured on Orin so the narrowing did not cost the point of the cache: 6.55 s/scene with it, 17.5 s without.

crates/apxinf-model/src/builtin.rs
qhy991: No reason — `qwen_drive-cuda` had no users anywhere in the tree. Removed in `5345d56`; only `qwen_drive` is registered now.

crates/apxinf-model/src/qwen_drive/mod.rs
qhy991: Agreed and done in `5345d56`. `perception_{depth,fpn,view}`, the cuDNN handle and convolution wrappers, pooling, group norm, batch-norm-relu, global mean and the NCHW/NHWC layout conversions are all out of the diff and kept under `devlocal/qwen-drive/perception/` for whoever connects the BEV path.

Sweeping for callers while removing them turned up more than the perception link: `nchw_to_nhwc_bf16`, `nhwc_to_nchw_bf16`, `batch_relu_bf16` and `global_mean` had no caller at all, not even a pending one.

VQA, direct planning and reasoning planning are unaffected — all three still verified on all three boards.

crates/apxinf-cuda/src/kernels/gemm/packed.rs
qhy991: Agreed, and removed in `5345d56` rather than kept switched off. Our own measurement is the argument against shipping it: the format is bit-exact, but its GEMV loses 2.7x below n=36864, which is where every decode GEMV in this model lives. `probes/steps.cu` put the whole difference on the instruction count — 608 µs for the three exponent planes against 217 for BF16 — so it is not a tuning problem.

`packed.rs`, `ffi/packed.rs`, `packed_bf16_adapter.cu` and the diagnostic GEMV beside them are under `devlocal/qwen-drive/`.

crates/apxinf-cuda/src/kernels/linear_attention.rs
qhy991: Correct — four layers, no caller. Removed in `5345d56`: the Rust wrapper, the FFI declaration, the adapter entry point and the kernel.

crates/apxinf-cuda/src/kernels/linear_attention.rs
qhy991: You are right, it is a reshape and nothing else. Fixed in `5345d56`: the call site in `vision.rs` uses `Tensor::reshape`, which on a GPU tensor rewrites the metadata and copies nothing, and the four layers of `merge_rows` are gone.

Checked that it changes no numbers — the precision probe is byte-identical across the change, all three modes, all four scenes.

doc/jetson-roofline.md
qhy991: Done in `5345d56`. Seven of the eight went to `devlocal/qwen-drive/reports/`: `RESULTS.md`, `THOR-ROOFLINE.md`, `jetson-roofline.md`, `kersor-qwen-drive-4090.md`, `qwen-drive-divergence.md`, `qwen-drive-integration-status.md`, `qwen-drive-performance.md`.

`doc/qwen-drive-devices.md` stays, cut from 407 lines to 101, and holds only what is not derivable from the code: which of the three layers a per-device decision belongs in (compile-time cfg, the runtime policy table, the tuning store), what the constants currently are, and which instrument decides a correctness question. Happy to drop that one too if you would rather `doc/` took nothing at all.

crates/apxinf-model/src/qwen_drive/vision.rs
qhy991: You are right, and it is worse than noisy — several of those lines carry prompt and generated token ids, which is request content leaving the process because a probe was left in.

Fixed in `5345d56`. All 27 now go through one `qdiag!` macro whose body is the `diagnostics_enabled()` check. A macro rather than 27 `if`s so that the switch is the only way to reach any of them and a new one cannot be added without it. Verified on both Jetsons: a normal request now writes exactly one line to stderr, and that line is `main`'s existing `CUDA 0: …` device banner in `backend.rs`, not this model.

crates/apxinf-model/src/builtin.rs
qhy991: Both predate this PR — they are lines 18-19 of `builtin.rs` on `main` — so I have left them alone, but for the record they are not redundant:

- `qwen3_vl` is the `model_type` string HuggingFace configs carry, and the detection path asserts on it (`crates/apxinf-model/tests/unified_llm_input.rs:344`, `python/apxinf/tests/test_checkpoint_layout.py:551`).
- `qwen3vl` is the bare spelling the CLI accepts (`src/main.rs:206` matches on both).

So dropping either would break something. Worth a follow-up to normalise the spelling in one place if you want, but it seemed out of scope here.

crates/apxinf-model/src/qwen_drive/general.rs
```diff
@@ -0,0 +1,2168 @@
...
+                    self.diag_digest.push(format!(
+                        "[qwen_drive] scatter_rows r0_l2={:.4} r0_first4={:?} rimg={} rimg_l2={:.4} rimg_first4={:?}",
+                        r0_l2, r0_first4, img_row, rimg_l2, rimg_first4));
+                }
+            }
+            // TEMP-DIAG (implement_r5, successor synthesis_r4 bundle 3 L6): pre-merger frame-0
+            // localization leg -- one ~7.1MB readback of the vis.pre_merger frame-0 slice;
+            // degenerate pre_merger => tower, diverse => merger chain; digest-routed; revert in
+            // the acceptance-bound revision.
+            {
```
ChenxiHe231: Gate remaining diagnostic readbacks

`qdiag!` only disables printing; this block still performs GPU-to-CPU copies during every normal multimodal prefill. The following `vis_premerge` block copies about 7.1 MB and the tail-slab probe at line 1597 also synchronizes the CUDA stream. Wrap all these readback/digest blocks in `diagnostics_enabled()` or remove the temporary diagnostics.

python/apxinf/apxinf/policies/impls/qwen_drive.py
```diff
@@ -0,0 +1,708 @@
...
+                acceleration,
+                ego,
+                nav_command,
+                noise_array,
+                self.max_new_tokens,
+                self.min_new_tokens,
+                terminators,
+                self.im_end_id,
+                self.newline_ids,
+                None,
```
ChenxiHe231: Forward the configured planning step count

`from_pretrained(num_steps=...)` stores this override and reports it in metadata, but both planning calls pass `None`, so Rust falls back to the checkpoint value. Pass `self.num_steps` to both `plan_reasoning` here and `plan_direct` at line 675.

qhy991:
## Post-merge precision review and request for maintainer review — `d9b700c`

The Thor and RTX 4090 reruns are complete. I am requesting another review of this head, with the remaining precision limitations stated explicitly below. This is a review handoff; the current evidence does not establish a completely green correctness gate.

### Results on the current head

| Device | CUDA tests | Additional checks | Alternating performance |
|---|---|---|---|
| Thor, sm_110 | 141 passed, 0 failed | 5 build/config tests passed; all 16 captured token/trajectory arrays matched the archived premerge binary byte for byte | 1.000932× |
| RTX 4090, sm_89 | 135 passed, 3 failed | All four fp64 operator oracles passed; full-model precision qualification remains open | 1.000338× |

Performance uses four public scenes, BF16, 64 generated tokens per request, one warmup and three measured requests per scene, three alternating premerge/merged rounds. Ratios are the geometric mean of the 12 paired scene medians. Neither result supports a material performance change.

The old Thor script silently failed to switch wheels: `pip` rejects the renamed files `premerge.whl` and `merged.whl`. The new rerun used isolated native libraries and checked the module actually loaded by each probe process. The table above uses that corrected rerun.

### RTX 4090 precision: controlled result, not an attributed merge regression

An earlier instrumented comparison returned VQA reference agreement `0.803951` for the archived premerge binary and `0.579562` for the merged binary. However, the same premerge binary also returned different results under a different probe wrapper. That comparison alone cannot identify a regression introduced by this merge.

I then held the probe, late native-module attestation, complete-output capture, inputs, CUDA libraries, and allocation-cache setting fixed, and varied only output clearing:

| Output-buffer setting | Premerge agreement to reference | Merged agreement to reference | Complete premerge/merged VQA token comparison |
|---|---:|---:|---|
| `APXINF_CUDA_SKIP_OUTPUT_ZERO=1` | 0.579562, 1370 compared tokens | 0.579562, 1370 compared tokens | **4/4 scenes exactly equal**, including lengths `[319, 93, 565, 399]` |
| Variable **unset**, restoring output clearing | 0.510901, 1376 compared tokens | 0.803951, 1316 compared tokens | **1/4 scenes equal**; first differences in scenes 0/2/3 at token indices 57/46/164 |

`APXINF_CUDA_ALLOC_CACHE=1` remained fixed. Both binaries loaded CUDA 12.3 / cuBLAS 12.3.2.9. Setting the clearing variable to the string `"0"` would still select dirty buffers, so the zero-fill control **unsets** it, matching [`output_fill`](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-cuda/src/workspace.rs#L269).

This establishes sensitivity to output clearing in these runs. It does **not** distinguish an initialization/lifetime/ordering problem from another numerical-path change, and it does not justify calling the discrepancy harmless BF16 noise. The reference-agreement ratios also use different generated lengths; they are not semantic-quality scores. The new controls cover VQA; the earlier direct/reasoning trajectory differences have not been requalified by these controls.

Binary identities used throughout:

```text
premerge  4c150168f6d375b1eee2e267b281f186f3c5e041c9c86e696d63421b5fa403aa
merged    e0d46f0caea6a203100b545167b9d80bf89bb4bdbe596dbc311bea987a91a7eb
```

The archived premerge binary was not freshly rebuilt from a clean parent checkout in this review. The candidate runtime sources were checked against `d9b700c`; this is an artifact-to-artifact comparison with the stated provenance limit.

### Source-review findings relevant to interpreting the measurements

1. **The generic head-64 tile change is not an established cause.** The default Qwen-Drive vision path calls [`fa2_attention_splitkv`](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-cuda/src/kernels/attention.rs#L2440), reaching the dedicated [`64×256` head-64 specialization](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-cuda/adapters/fa2_head64_adapter.cu#L9). That adapter is unchanged across the merge. The generic `128×64 → 128×128` source diff therefore cannot, by itself, explain these default-path outputs.
2. **The oracle's printed route label is misleading.** [`operators.rs:693`](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-cuda/src/tests/operators.rs#L693) prints `composed` when `APXINF_VISION_FA2` is absent, whereas the production predicate enables the specialized FA2 route by default. Also, setting that variable to `0` on sm_89 falls through to generic FA2, because the composed arm is compiled only under `not(apxinf_fa2_sm80)`. Do not use those labels or that switch alone as evidence of a composed-vs-FA2 numerical comparison.

The four fp64 oracles passed on both boards, but that does not close the model-output sensitivity above. A useful next investigation is to bind a clean parent build and compare fixed inputs/intermediates at the first divergent endpoint, with buffer initialization and stream ordering checked explicitly.

### Existing failures and open review items

RTX 4090 still fails:

- `dynamic_fp8_row_channel_scales_match_bf16_reference`: the implementation requires an SM100-family native backend.
- `fp8_identity_gemm_runs_on_device` and `fp8_large_k_gemm_matches_quantized_cpu_reference`: `CUBLAS_STATUS_NOT_SUPPORTED` (status 15).

Those test bodies and the FP8 GEMM implementation did not change in the latest main merge. Without a freshly built parent test run, I am not relabeling them as proven pre-existing failures.

There are currently **12 review threads: 9 resolved, 3 open**. The original round contained 10 threads, including one follow-up comment; it was not 11 independent threads. The open items are:

- [The pre-existing `qwen3_vl` / `qwen3vl` aliases](https://github.com/infinigence/ApxInf/pull/78#discussion_r4044392292).
- [Remaining diagnostic GPU readbacks](https://github.com/infinigence/ApxInf/pull/78#discussion_r4055999271): confirmed by source inspection; guarding printing does not guard the readbacks.
- [The planning `num_steps` override is not forwarded](https://github.com/infinigence/ApxInf/pull/78#discussion_r4056011288): confirmed at both planning call sites. The default-step probe does not validate this override.

No production code was changed as part of this review, and these items remain open for maintainer assessment. Complete logs, commands, native-library identities, token arrays, and the four-arm control receipt are preserved under `devlocal/qwen-drive/precision-review-20260920/` on the test workspace; the earlier three-mode and performance evidence remains under `devlocal/qwen-drive/revalidation-20260920/`. The completed control run is `gpuq-097f42a8d701`; the earlier unsuitable vision-route experiment was cancelled and is excluded from the results.

@wong-1994 @ChenxiHe231 — could you please review `d9b700c` again, including the remaining precision questions and the three open threads above?

ChenxiHe231:
@qhy991 Thanks for the controlled rerun. The result is useful, but it does not close the correctness gate. Changing only output-buffer initialization changes complete 4090 model outputs, which contradicts the source assumption that all output_buffer consumers tolerate dirty memory. Please locate the first divergent operator and verify it under zero, dirty, and poison fills before treating this path as qualified. A clean build of the PR parent is also needed, and the direct/reasoning trajectory paths remain unqualified. The diagnostic readbacks, ignored num_steps override, and zero layers_per_kv validation issue also remain unresolved.

wong-1994:
Re-review of d9b700c: I would hold approval until the correctness qualification and the concrete implementation issues below are addressed. This review is based on source inspection and the reported device experiments; I have not independently rerun the GPU tests.

1. **Merge blocker: unexplained RTX 4090 output sensitivity to buffer clearing.** The controlled results show that changing output initialization changes complete model outputs. This leaves the full-overwrite assumption behind output_buffer unqualified; it does not yet establish whether the cause is an initialization, lifetime, ordering, or numerical-path issue. Please rebuild the parent from a clean checkout, locate the first divergent operator with identical inputs, and compare zero/dirty/poison fills. Requalify VQA, direct planning, and reasoning planning after the fix. Passing the four operator oracles does not close this full-model gate. See the [reported controls](https://github.com/infinigence/ApxInf/pull/78#issuecomment-5748080329) and [output-buffer policy](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-cuda/src/workspace.rs#L247).

2. **P2: gate the entire diagnostic readback blocks, not just their printing.** Confirming the existing review thread: the scatter-row, pre-merger, and tail-slab probes still run with diagnostics disabled (general.rs lines 1527, 1552, and 1597). Each to_cpu call reaches cudaDeviceSynchronize before the device-to-host copy; normal inference therefore still pays device-wide synchronization, transfer, and CPU statistics costs. Guard the complete probe blocks with diagnostics_enabled(), or remove them. [Readbacks](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-model/src/qwen_drive/general.rs#L1527), [synchronization](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-cuda/src/transfers.rs#L80).

3. **P2: forward the configured planning step count.** Confirming the existing thread: both Python planning calls pass None, although self.num_steps is accepted, stored, and reported. Rust then falls back to the checkpoint default. Pass self.num_steps to both calls and add coverage showing a non-default value reaches the native interface. [Reasoning call](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/python/apxinf/apxinf/policies/impls/qwen_drive.py#L652), [direct call](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/python/apxinf/apxinf/policies/impls/qwen_drive.py#L675).

4. **P2: reject zero layers_per_kv before division or remainder.** JSON parsing accepts zero, but PlanningExpertConfig::validate immediately evaluates n_layers % layers_per_kv; num_kv_sources also divides by it. Invalid configuration consequently panics instead of returning a configuration error. Validate layers_per_kv > 0 before these operations and cover the zero-value loading case. [Validation](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-model/src/qwen_drive/config.rs#L161).

5. **Make the vision oracle report its actual dispatch route.** operators.rs labels an unset APXINF_VISION_FA2 as composed, while the production predicate enables the specialized path by default. Conversely, the string 0 is labeled fa2-head64 by the oracle, although it disables the specialized route; on sm_89 this does not establish a composed-path comparison either. Report the selected implementation rather than inferring it from variable presence, so subsequent numerical comparisons identify what actually ran. [Oracle label](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-cuda/src/tests/operators.rs#L693), [production predicate](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-cuda/src/kernels/attention.rs#L2374).

6. **Dependency direction: weights should not depend on the executor.** device_weights::projection calls general::tuned_projection to decide physical weight layout. This conflicts with the existing model-layer dependency rule. Resolve the layout choice during loading/construction and pass the same explicit choice to weight packing and computation. This can be a small local change; a full PR 50-style model/runner migration is not a prerequisite for this PR. [Call site](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-model/src/qwen_drive/device_weights.rs#L109).

The absence of whole-model CUDA Graph capture or further fusion is not itself a merge blocker. The immediate priority is the unresolved correctness evidence and the fixes above.

wong-1994:
One additional P2 finding for d9b700c: **do not automatically load a planning expert through the text-only registry entry point.**

In [builtin.rs lines 92–101](https://github.com/infinigence/ApxInf/blob/d9b700c1d7521c453f446da41f7ebd2b0256bba2/crates/apxinf-model/src/builtin.rs#L92), load_qwen_drive detects an adjacent planner-sft directory and passes it to load_with_backend, but returns LoadedModel::Text(Box<dyn LlmTrait>). That returned interface cannot invoke plan_direct or plan_reasoning. A VQA/text caller therefore pays for expert weights it cannot use through this interface; an invalid or incomplete adjacent planner checkpoint can also fail an otherwise valid VLM load.

Please load only the VLM through the text entry point and load expert weights explicitly through a planning-capable entry point. Directory presence alone should not select extra deployment capabilities. Cover the case where a valid VLM has an unusable planner-sft directory beside it: text loading should remain independent of that planner, while an explicit planning load should validate it and report its error.

This can be fixed locally and does not require the full PR 50-style model/runner restructuring.

crates/apxinf-model/src/qwen_drive/general.rs
qhy991: Fixed in a1d5db9e9e450f775265fca7d40dfed5a2180730.

The scatter readbacks, ~7.1 MB vision premerge readback, and tail-slab readback are now inside `diagnostics_enabled()`. The adjacent digest formatting, tail-token probe setup and diagnostic segment map are guarded by the same switch, so disabling `qdiag!` no longer leaves their work on the normal path. Explicitly requested trace capture remains opt-in.

Validated on RTX 4090 with diagnostics unset: all three supported modes run, and the complete four-scene token/trajectory captures are byte-identical with output clearing enabled versus disabled. This is a stability check, not a claim of complete Hugging Face parity. The separate FA2 reference-kernel symbol collision discovered during the precision investigation is fixed in 264cd4c.

python/apxinf/apxinf/policies/impls/qwen_drive.py
qhy991: Fixed in a1d5db9e9e450f775265fca7d40dfed5a2180730. Both `plan_direct` and `plan_reasoning` now receive `self.num_steps`, the same resolved value reported in metadata.

Added four focused policy tests: direct/reasoning planning, each with checkpoint-default steps and an explicit override. They execute the real lightweight preprocessing and assert that the native call receives exactly the metadata step count. All four pass; native end-to-end runs of both planning modes also complete on RTX 4090.

qhy991:
@wong-1994 @ChenxiHe231 The follow-up fixes and current-main integration are ready for another review. Head is **d12390a21b366085c59df50fb75f3c67f81d8d32**, including main **a1a37af**. This supersedes the open 4090 failure status in my earlier comment, within the limits below.

### Changes

- **a1d5db9:** guard remaining diagnostic GPU readbacks/digest work, forward `self.num_steps` to both planning calls, and add four default/override policy tests. Both code-review threads have a detailed reply and are resolved.
- **9f9907f:** correct the physical TN layout for legacy FP8 cuBLASLt execution; staging belongs to the calling stream/workspace, not the shared shape plan. F16/BF16 outputs and same-shape plan reuse with changed weights are covered. Dynamic rowwise FP8 remains explicitly unsupported on 4090; its test now checks that contract instead of unwrapping an unsupported operation.
- **264cd4c:** separate dedicated reference-math FA2 kernel entrypoints from generic fast-math entrypoints. The compiled objects had 10 duplicate weak kernel host stubs; the final audit reports 10 reference stubs, 72 generic stubs, **0 duplicates**. This is a concrete instance of the template-stub linking hazard described in the [NVIDIA nvcc documentation](https://docs.nvidia.com/cuda/archive/12.8.0/cuda-compiler-driver-nvcc/index.html#static-global-template-stub).
- **d12390a:** integrate #77/#50/#80 without undoing their module ownership or reduction fixes. Adapt PR-specific cached LayerNorm and two softmax kernels to the scratch-reuse barrier contract; use negative infinity for inactive max lanes and extend the regression with all-negative inputs.

### Current-head validation

- RTX 4090 / SM89 / CUDA 12.3: **143 CUDA tests + 5 architecture tests pass**, zero failed/ignored.
- Policy/registry/layering CPU tests: **33 pass, 2 native-dependent tests skipped**; all four new Qwen planning tests pass. Model-family and PI0.5 boundary checks pass.
- All three supported modes, all four scenes: **16 complete token/trajectory arrays are byte-identical** with output clearing enabled versus disabled.
- An uninstrumented scene-1, 32-token control gives identical complete tokens with cache on, cache off, EAGER module loading and early native import. No `CUBLAS_WORKSPACE_CONFIG` workaround.
- Scratch reuse / softmax / SDPA reference regression: Compute Sanitizer racecheck reports **0 errors, 0 warnings on both 4090 and Thor**. Thor also compiles the changed FP8/reference-FA2/custom adapters.
- Verified 2,953 runtime source files against the final Git tree. The actually loaded 4090 native library is SHA256 `05798355a6f27bded26492668f71eac272c5035f5f956a646c865a92b949ad28`.

### Performance and limits

Three alternating baseline/fixed rounds, four scenes, BF16, 64 tokens, warmup 1 / repeat 3: geomean **0.99161468×** versus the archived d9b700c library, approximately **0.85% slower**. Eleven of twelve scene-round ratios are 0.9935–1.0084; one is 0.90284. All samples are retained. This is a measurement, **not an acceleration claim or a pass of the original 5% optimization gate**. The baseline is an archived binary, not a freshly rebuilt clean parent.

HF full-model parity remains a separate gap: the fixed-reference VQA token agreement is 0.803951, direct trajectory p99 is 0.16113287, and reasoning token agreement is 0.926471. These measurements do not justify calling every numerical difference solved. Perception remains unsupported. Current-head Thor evidence is compilation and the focused operator regression, not a full-model rerun; Orin was not rerun on this head.

There are now **12 threads, 11 resolved**. I left the pre-existing `qwen3_vl` / `qwen3vl` compatibility discussion open for the maintainers, preserving both consumers. The PR is open, non-draft and mergeable; I have not merged it. Detailed logs, raw numerical captures and private control scripts remain under `devlocal/qwen-drive/`, outside the review diff as requested.

qhy991:
@wong-1994 @ChenxiHe231 Follow-up to the [six-point re-review](https://github.com/infinigence/ApxInf/pull/78#issuecomment-5748251587) and the [text-registry planner finding](https://github.com/infinigence/ApxInf/pull/78#issuecomment-5748273489): the new implementation fixes are pushed at **5d7d78565ce23aaf38e756e1e95d6b9a63110057**. I had missed the additional findings in general PR comments when counting resolved inline threads; that count was not a complete review checklist.

### Implementation fixes

**ed77678** addresses the three remaining concrete implementation issues:

1. `PlanningExpertConfig::validate` rejects zero `layers_per_kv` before remainder. Top-level config validation validates the expert before computing its cache count. JSON loading and manually changed config regressions pass, including the non-divisible case.
2. Projection layout is resolved once at model construction, passed explicitly to packing, and retained with the device weights. Execution reads that same value. `device_weights` no longer calls `general::tuned_projection`; no full model/runner migration or new backend abstraction was added. Both BF16 packing layouts are tested.
3. The text registry passes `None` for the planner. A real CUDA/checkpoint test creates an invalid adjacent `planner-sft`: `AutoModel::load_model` succeeds as Text, while explicitly requesting that planner returns a planner-loading error. This is an actual native integration test, not a mocked directory-selection test.

The diagnostic-readback, planning-step forwarding and oracle-route-label items were already fixed in **a1d5db9** and remain in this head.

**5d7d785** extends the maintained standalone reduction regression with cached LayerNorm at 512/1024/2048 columns, repeated zero/pattern/poison outputs and a CPU reference. The implementation barrier was already included when integrating #80 in d12390a; this commit adds direct regression coverage for that PR-specific kernel.

### Clean parent and all-mode fill controls

Rebuilt **7a630a3787ec24d0060f544dd505ed2253d54ddd**, the first parent of the originally reviewed d9b700c main merge, from a clean worktree with a new empty Cargo target directory. Verified 2,856 runtime source files against its Git tree; its lockfile remained unchanged after building. This is not the archived parent library used previously.

Both parent and candidate ran the same existing precision probe, four scenes, fixed noise and all three supported modes in fresh zero/dirty/poison processes, with allocation cache on and no diagnostic readbacks, fallback model, launch-blocking or workspace workaround. Actual native module and CUDA-library identities were attested after each run.

| Result | Clean parent | Current implementation |
|---|---:|---:|
| Byte-identical zero-vs-dirty / zero-vs-poison comparisons | **6/32** | **32/32** |
| All three modes / four scenes complete | yes | yes |
| Match all 16 outputs of the prior fixed d12390a head | not claimed | **16/16** |

The clean parent reproduces the sensitivity: VQA agreement to the fixed reference is 0.510901 for zero and 0.579562 for dirty. Direct/reasoning trajectories also differ between fills. The candidate's three fill arms all report VQA 0.803951, direct p99 0.16113287 and reasoning token agreement 0.926471. These are reference-difference statistics, not full HF parity.

Native SHA256 identities:

```text
clean parent ad11d50e18f64c9eea1c2e99024c8df079500eb904a4f82cc7fe373b56e029ac
candidate    172d90d757cd28ac7d86ccb052170d006e3b6ce70c935e14bf89db540fbb0964
```

Current-head checks also pass: 12 Qwen model unit tests, the real registry/planner integration test, module boundaries, and the extended Compute Sanitizer regression (0 errors, 0 warnings). The final head rebuild produces the same tested native bytes; its last commit changes only the standalone regression.

### What is still not closed

The cached LayerNorm counterfactual uses the exact same test program and inputs against parent versus fixed headers. Parent racecheck reports **9 error groups** at the mean-read / variance-write boundary and produces errors around 0.68–0.73; fixed headers give **0 race errors** and stable outputs. Parent happened to pass the small uninstrumented replay, so this is a demonstrated scheduling-sensitive operator defect, not proof that it was the first divergence in the historical full-model run.

I also captured the first real vision cached LayerNorm: all **12,509,184 input and output elements**, weights and bias were identical across parent-zero, parent-dirty and candidate-zero. The 64-token capture run differed from the longer full precision probe, but a matched 64-token, same-script control gives the same parent-zero prefix without instrumentation. This corrects my earlier wording that the capture itself changed that prefix: the cross-probe comparison did not isolate that cause. The first captured LayerNorm boundary has no observed divergence and **does not satisfy the requested first-divergent-operator localization**. I am explicitly leaving that part of the merge-blocker open, rather than attributing every historical discrepancy to either this race or the FA2 weak-symbol collision.

Please re-review the implementation corrections and added evidence. This is not a request to waive that remaining causal-localization requirement. No new performance claim, Thor/Orin full-model qualification or PR merge is made. Private probes/captures remain under `devlocal/qwen-drive/`, outside the product diff.

qhy991:
Merged current main ab498b0 into this PR in f4bbc0145a2494cf1015b4012783f52f9770fda7. The only conflict was doc/adding-new-kernels.md: retained the new upstream cuda-new guide from #85; the legacy toolkit-qualified tuning-store explanation remains in doc/qwen-drive-devices.md. All prior Qwen-Drive fixes and its legacy CUDA/model/Python/config sources are unchanged across this merge. Validation: no unresolved paths, diff whitespace check clean, model-family and PI0.5 boundary checks pass, and 10 CPU Qwen-Drive unit tests pass. No GPU rerun or new numerical qualification is claimed. The previously documented causal-localization review requirement remains open; this update only resolves the merge conflict.

PR #86, opened 2026-09-20
feat(pi0-fast): add FP8 E4M3 decode with measured activation calibration and a pruned LM head

Kimho666:
π0-FAST gains an FP8 E4M3 decode path on top of the existing BF16 one, gated behind
  precision="fp8" and a measured activation-calibration profile. Both paths share the
  same architecture code, KV-cache loop and BF16 tails; only the projection dispatch and
  the LM head differ.

  ## Optimizations

  FP8 E4M3 projections. Every linear projection runs as E4M3 x E4M3 with per-tensor
  weight scales and per-site measured activation scales, accumulating in FP32 and writing
  BF16, so norms, RoPE, attention, the KV cache, the argmax and the tied embedding lookup
  are reused unchanged. Decode step: 18.7 ms -> 10.6 ms on Jetson AGX Thor (sm_110, 20 SMs).

  Pruned LM head. The head is the largest read in a decode step (1.05 GB of 5.01 GB in
  BF16) and only its argmax is consumed, so it keeps the 2048-wide action window at the
  vocabulary tail plus action_head_extra_tokens — the "Action: " prefix and the "|"
  terminator, which are nowhere near the tail and would otherwise change the first three
  steps and make the stop condition unreachable. 2112 of 257152 columns are retained
  (padded to a multiple of 64 because cuBLASLt rejects an unaligned E4M3 leading
  dimension), cutting the per-step head read from 1.05 GB to 8.6 MB in BF16 and from
  526 MB to 4.3 MB in FP8.

  CUTLASS wide-tile FP8 tactics for the BF16-output GEMM. The FP8-to-BF16 entry point
  previously resolved a tuning plan and then ignored it, always dispatching to cuBLASLt.
  It now executes the selected tactic and falls back to the vendor path on failure, which
  lets the wide decode shapes use the CUTLASS schedules: Gate+Up 428.3 -> 363.4 µs (1.17x)
  and Down 241.1 -> 199.4 µs (1.21x).

  No uncalibrated default. A uniform scale does not preserve the FAST token stream: the
  decode stops emitting its terminator and spends its whole budget on tokens the policy
  discards (LIBERO-10: 0/6 before the run was abandoned). FP8 therefore refuses to load
  without a profile and the uncalibrated path is reachable only through an explicit
  APXINF_PI0FAST_FP8_ACTIVATION_SCALE, which warns.

  ## Implementation

  - pi0fast/fp8_executor.rs, fp8_runtime.rs, fp8_weights.rs, static_fp8_weights.rs:
    the E4M3 weights, per-site scales and the layer/decode-loop mirror of the BF16
    executor. fp8_projection quantizes its BF16 activation under the calibrated scale and
    dispatches to gemm::fp8_bf16.

  - pi0fast/fp8_calibration.rs: the ordered 182-site plan derived from the config, the
    profile schema (apxinf.pi0fast.fp8-calibration.v1), full validation (schema, family,
    site coverage, amax/scale consistency, synthetic-vs-production labelling,
    quantization contract) and a cross-language sha256:... checkpoint identity. Strict
    coverage means a profile from a different config is rejected rather than partially
    applied; a checkpoint mismatch is reported as a warning, as in PI0.5.

  - pi0fast/calibration.rs: a BF16 Bf16ActivationObserver keyed on weight device
    pointers, so it records exactly the operands the FP8 runtime will quantize without
    threading anything through the layer signatures. The capture drives the real inference
    path — same prompt, same prefill, same autoregressive steps — and honours the FAST
    terminator, because a free-running decode records activations no rollout reaches.

  - vla_runtime.rs: one RuntimeVariant front end over the two runtimes, auto resolving
    to BF16, and resolve_fp8_scales with the order explicit calibration= -> the
    checkpoint's own calibration.json -> explicit environment variable -> error.

  - config.rs: action_vocab_size, action_head_extra_tokens and action_head_columns(),
    validated against the vocabulary at construction.

  - static_bf16_weights.rs: the LM-head host transpose became a gather-and-transpose over
    the retained columns, which also removes a 2.1 GB float32 intermediate at load time.

  - Argmax: a dedicated argmax_remap_bf16_kernel / apxinf_argmax_remap_bf16 /
    sampling::argmax_bf16_remapped_into writes remap[argmax], so the decode loop, the
    tied embedding lookup and the caller's stop_token keep speaking global token ids. The
    full-vocabulary operator is untouched and byte-identical to before.

  - apxinf-cuda: Fp8Gemm gained an output-element template parameter, fp8_gemm_bf16
    exposes the same eight wide-tile tactics as fp8_gemm_f16, and the CUTLASS provider
    accepts GemmOp::Fp8Bf16.

  - Python: Pi0FastPolicy.from_pretrained(..., calibration=, autotune=),
    calibration_plan() / collect_calibration(), the _calibrate_tokens_rgb binding, and
    scripts/calibrate_pi0fast.py with its profile/data helpers. Also fixes
  ## Results (Jetson AGX Thor, sm_110, 20 SMs, pruned head, shipped tactic DB)

                                            BF16       FP8 E4M3
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━  ━━━━━━━━━━
   per autoregressive token                 18.7 ms    10.6 ms
  ───────────────────────────────────────  ─────────  ──────────
   LIBERO-10, 10 tasks x 1 trial, seed 7    6/10       4/10
  ───────────────────────────────────────  ─────────  ──────────
   mean model time per LIBERO-10 call       504 ms     313 ms

  The FP8 path is 1.61x faster end to end and keeps the terminating token stream; it does
  not yet match BF16 accuracy, so it is a speed path with its own accuracy budget rather
  than a drop-in replacement. Both paths were verified to reproduce the pre-change token
  streams bit for bit (LIBERO-10 task 0 checksums 11.331596374511719 BF16 and
  11.179217338562012 FP8).

  ## Notes

  The tactic ids in configs/tuning/.../tactics.json are measured on one board only. The
  FP8-to-BF16 projections have their own records (op=fp8_bf16) and likewise ship only a
  Thor entry, so another board keeps the cuBLASLt default until it runs a local autotune.


crates/apxinf-model/src/pi0fast/README.md
```diff
@@ -131,6 +157,22 @@ pure streaming read: the decode step is at the memory roofline. `nsys` already
 shows 97.9% GPU busy, so no launch overhead remains to remove, and the only
 headroom left is reading *fewer* bytes.
 
+### Jetson AGX Thor
```
zealoct, line 160: should we add the results to main README?

Kimho666: No, I am still optimizing the performance.

ChenxiHe231 (approved):
Looks good to me

Kimho666:
@ChenxiHe231  ok, I use the implementation of dynamic cublaslt and cutlass selection.

PR #88, opened 2026-09-21
Pi05 with cuda new

ChenxiHe231:


wong-1994 (changes requested):
# Review: architecture, lifecycle, and evidence

Reviewed at `9ea4bbb5` against merge-base `ab498b08` (191 files, +16287 −2204).

Before the findings, what this PR gets right, because it matters for how the rest should be read:

- **CUDA Graph capture safety is genuinely stronger than before.** `ops::with_session` is inside the `capture_cuda_graph` closure so the arena reset happens after `cudaStreamBeginCapture`; `may_prepare_native_resources()` is false while capturing (`workspace.rs:235-239`); GEMM and attention hard-error on a cache miss instead of tuning (`gemm_execution.rs:113-121`, `attention/execution.rs:68-76`); `execute()` skips `ctx.synchronize()` under capture; every stateless launch calls `validate_capture_target`. I could not find a capture-unsafe call reachable in the captured region.
- **Layer direction is clean.** No file under `pi05/model/` references `model_runner`, `ExecutionSession` or `CapturedGraph`. The runner does not recreate model mathematics.
- **`backend.rs` is still a seam**, not a provider abstraction, and `src/pi05/` contains no `unsafe` and no raw FFI.
- **The destructive commits did not survive the rebase.** `git diff ab498b08 -- crates/apxinf-model/src/{llama,walloss,gr00t,qwen3vl}` is empty and the legacy crate still exists. The branch history contains a delete-and-restore of ~1.3M lines, which makes commit-by-commit review impractical, but the net tree is fine on this axis.
- **`baseline.md` and `pi05-cuda-regression.md` honestly downgrade previously overstated fixture claims** rather than carrying them forward.

Layer order, residual order, norm epsilon and scale-application order in the three Blocks files are line-for-line equivalent to merge-base. The migration is mechanically faithful. The problems below are about ownership, fallback, and evidence.

---

## Blocking

### B1. `CapturedGraph` drop order can destroy the native runtime before its executions

`CudaContext::drop` calls `apxinf_runtime_destroy` (`cuda-new/src/context.rs:136-140`). GEMM/attention `Execution` objects are created from that runtime and their `Drop` calls `apxinf_gemm_destroy` / `apxinf_attention_destroy` (`gemm_execution.rs:97-102`, `attention/execution.rs:60-65`), but they hold only `Arc<CudaStream>` — never `Arc<CudaContext>`. `ExecutionSessionInner` (`workspace.rs:74-79`) holds no `Arc<Context>` either.

Field order in `CapturedGraph` (`model_runner/prepare.rs:27-30`) is:

```rust
backend: Arc<Context>,
_fixed: Box<dyn std::any::Any>,
session: ExecutionSession,
```

Rust drops fields in declaration order, so `backend` is released before `session`. When the prepared handle holds the last `Arc<Context>` — which the lifecycle doc explicitly allows, since a plan may outlive the runner — the runtime is destroyed while its executions are still alive, and the subsequent `apxinf_gemm_destroy` runs against a freed runtime. `EagerInputs` (`runner.rs:63-69`) holds no `Arc<Context>` at all.

This could not happen pre-PR: the legacy `GraphWorkspace` held only device buffers (`cudaFree` is context-independent) and native handles lived inside the shared `Arc<CudaContext>`.

Please make the ownership explicit rather than relying on declaration order — store an `Arc<Context>` in `ExecutionSession` (and in `EagerInputs`). Field reordering alone would fix the symptom but leaves the next refactor free to reintroduce it.

### B2. `packed_qkv_attention` has no fallback, so its admission rule is duplicated across languages

`fp8_static.rs:815-819` branches on hardware in the model layer:

```rust
let use_packed_qkv = ctx.caps().arch_family == apxinf_cuda_new::CudaArchFamily::Sm100
    && patches_per_view == 256 && heads == 16 && head_dim == 72
    && weights.qkv.bias.is_some();
```

This is necessary today only because the operator cannot fall back. `packed_qkv_entries` (`native/adapters/attention/candidates.cpp:177-184`) contains exactly one implementation, gated on `kDeviceFeatureCutlassSm100`, with `supports_fa2_packed_qkv` (`candidates.cpp:60-66`) hardcoding `query_tokens == 256 && query_heads == 16 && kv_heads == 16 && head_dim == 72 && dtype == F16`. On anything else the call fails with no candidate, even though `dense_entries` lives in the same file.

So the same admission predicate now exists twice — once in C++, once in Rust — with nothing keeping them in sync, and an architecture identifier (`Sm100`) has entered the model layer. `fp8_static.rs:815` is also the only direct `apxinf_cuda_new::` import in the model crate, bypassing the declared `backend.rs` seam.

Preferred fix: make `PACKED_QKV` fall back to split + dense inside the framework, so the model layer calls one function and carries no arch branch. Acceptable alternative: expose a support query (`ops::attention_supports(&args) -> bool`) so the model asks instead of duplicating the predicate. Either way, please remove the `CudaArchFamily` import from `model/blocks/`.

### B3. No numerical parity evidence for the migrated path, while README keeps legacy-backend numbers

There is no test anywhere comparing migrated PI0.5 against the legacy backend or against torch fixtures. `grep apxinf_cuda:: crates/apxinf-model/src/pi05/` returns nothing, so an old/new comparison is not even constructible in-tree. The closest thing, `runner.rs:757 native_prepare_for_reuses_prepared_execution`, is `#[ignore]`-gated on CUDA plus `APXINF_PI05_TEST_CHECKPOINT`, feeds all-zero patches and noise, and asserts only run-to-run equality plus `is_finite()`. Torch-golden coverage was extended for exactly one new semantic (`FP8_UNIT_GEMM_BIAS_RESIDUAL`); the five migrated direct families (norm/pointwise/rope/gather/quantization) are checked only against hand-written CPU references.

Meanwhile `README.md:92-110` keeps every latency cell (Thor BF16 72.45 ms, FP8 41.16 ms, Orin 165.67 ms, 4090 31.38/21.92 ms) and `README.md:115+` keeps LIBERO-10 92.8%, unchanged, after the entire operator stack moved crates. This PR's own new gate, `doc/pi05-cuda-regression.md:220-225`, says a cell more than 3% slower "is a migration failure" and requires re-measurement on Thor and Orin. There is no CI in this repo (`.github/workflows` does not exist), so nothing catches this automatically.

Please either (a) re-measure and update the README tables per the gate this PR itself introduces, or (b) mark them as pre-migration numbers pending re-qualification. And please add at least one end-to-end numerical comparison for the migrated path — a captured-vs-torch-fixture check on a real checkpoint, not a zeros smoke test.

The cost of this gap is already visible in the branch: `git log -S` shows that `f727c8ed` ("restore non-causal action attention"), `849dadd5` ("slice FP8 adaptive norm modulation") and `3a07627e` ("preserve I32 accumulation for W8A8") all repair regressions introduced by this PR's own migration commit `e5d88850`, not pre-existing bugs. The causality regression survived ~55 commits. `GemmPolicy` and `query_start` do not exist at merge-base.

### B4. `check_kernel_architecture.sh` moved its coverage instead of extending it

`scripts/check_kernel_architecture.sh:7` introduces `cuda_root=crates/apxinf-cuda-new` and retargets all seven pre-existing rules to it. The legacy crate — which this same PR declares the global default for WallOSS/GR00T/Llama/Qwen3-VL/pi0fast (`doc/model-layer-architecture.md:76`) — is now covered by no architectural rule at all.

Reproduction from the PR tree:

```
$ ./scripts/check_kernel_architecture.sh
kernel architecture checks passed                        # exit 0

$ git show ab498b08:scripts/check_kernel_architecture.sh > scripts/_base.sh && chmod +x scripts/_base.sh && ./scripts/_base.sh
crates/apxinf-cuda/src/kernels/attention.rs:210: ... std::env::var_os("APXINF_PREFILL_SDPA_LEGACY")
crates/apxinf-cuda/src/kernels/attention.rs:450: ... std::env::var_os("APXINF_VISION_SDPA_LEGACY")
crates/apxinf-cuda/src/kernels/attention.rs:474: ... std::env::var_os("APXINF_VISION_SDPA_LEGACY")
crates/apxinf-cuda/src/kernels/attention.rs:1136: ... std::env::var_os("APXINF_DISABLE_FA2_SPLITKV")
kernel architecture violation: kernel execution paths must not read environment variables   # exit 1
```

To be precise and fair: **this violation is pre-existing on main**, not introduced here — `attention.rs` is byte-identical to `ab498b08`. The problem is that the PR removes the rule that reports it while simultaneously declaring that crate the default backend for five families. Note also that `doc/pi05-fa2-sm89-bf16-opt.md` deletes the `APXINF_DISABLE_FA2_SPLITKV` documentation while the env read itself survives.

Also removed or narrowed:
- Deleted: the `kernels/custom/*.cuh` legacy-header classification loop, and the `crates/apxinf-cuda/native` directory ban.
- Narrowed: `'crate::launch|launch::|mod launch'` → `'crate::launch'` (:117) — required because cuda-new adds `pub(crate) mod launch` in five op families; and the env-var rule gains `--exclude-dir=tests` (:56).

Credit where due: the four new checks (legacy crate must exist, PI0.5-only cuda-new dependency, flat stateless adapters with no candidate lifecycle) are real additions.

Please pick one and make it explicit: either the guard covers both roots, with the legacy rules retained for `crates/apxinf-cuda`, or the legacy crate is declared frozen and read-only in documentation. The current state claims it is the default backend while exempting it from every rule.

### B5. PI0.5 bypasses the registry, and the exception is locked in by a test and a guard

`builtin.rs:21` removes `pi05::register_builtin()`. `auto.rs:248` adds `uses_pi05_cuda_new(model_name)` matching the literals `"pi05" | "pi05-cuda"`, and `auto.rs:~303` early-returns `crate::pi05::load_with_cuda_new(path, device, options)` before `register_builtin_models()` and `create_backend()`. PI0.5 is no longer in `registry::list()`, and a family-neutral file now carries family-specific branching, a family-specific error string, and a cuda-new-only `CudaRecipeOptions` type.

I understand why: `ModelFactory = fn(&Path, Device, Arc<dyn Backend>, &LoadOptions) -> Result<LoadedModel>` (`registry.rs:13`) requires a constructed `Arc<dyn Backend>`, and cuda-new's `Context` does not implement `apxinf_core::Backend` (no `impl ... Backend for` anywhere in the crate), so the old `downcast_arc` route is unavailable.

What concerns me is not the workaround but that it is being cemented. `tests/cuda_backend_routing.rs:23` asserts that `auto.rs` source text contains the literal string `"crate::pi05::load_with_cuda_new(path, device, options)"`, and `check_kernel_architecture.sh` adds the same grep as an architectural rule. Anyone restoring registry-based loading must first delete a test and a guard.

Suggested minimum fix: change `ModelFactory` to take `Device` + `&LoadOptions` and let each factory construct whatever backend it needs (one added `create_backend(device)?` line per existing factory). PI0.5 returns to the registry and `auto.rs` needs no family branch. If you'd rather defer that, please at least drop the source-text test and the guard rule so the exception stays visibly temporary.

---

## Should fix

### S1. FP8 weight quantization regressed from GPU to scalar CPU at load time

`weights/fp8_static.rs:52` calls `quantize_e4m3_absmax`, which is a per-element host loop over `encode_e4m3` (`weights/fp8_static_calibration.rs:652,662-682`). The deleted CUDA fast path's own comment read: *"Quantizing billions of parameters with the scalar CPU E4M3 encoder is prohibitively slow on Jetson."* cuda-new exposes `ops::quantization` and it is used per-request by the W8A8 path, but load does not use it. This is a silent Stage-2 load-time regression, most visible on Thor.

### S2. `tuning_path` / `autotune` are now silently ignored for every non-PI0.5 family

`auto.rs` deletes the `configure_cuda_tuning` call and the function itself, replacing it with `cuda_recipe_options` (`auto.rs:248-290`), which only PI0.5 consumes (`pi05/load.rs:153-156`). After this PR, `apxinf_cuda::kernels::gemm::configure_tuning` has zero production callers. For GR00T, Llama, Qwen3-VL and pi0fast, `LoadOptions.tuning_path` and `LoadOptions.autotune` now do nothing — no error, no warning. Only WallOSS survives via its private `install_tuning_db` (`walloss/bf16_runtime.rs:654`).

Please either keep the legacy wiring for legacy families or make the unsupported combination a load-time error. Silently ignoring a user-supplied tuning path is the failure mode `doc/model-lifecycle.md:84-85` explicitly prohibits.

Related: `auto.rs` maps a user-specified `--tuning-path foo.json` to a sibling `foo.recipes` directory without parsing or reporting anything. Even accepting that old tactics databases cannot be converted, this should warn rather than silently redirect.

### S3. Eager output now aliases the session arena, and the contract text was not updated

`run_eager` (`runner.rs:163-178`) now runs under `ops::with_session`, so its output is sub-allocated from the arena (`context.rs:126` → `workspace.rs:348`) and is invalidated by the next run's `workspace.reset()` (`workspace.rs:194`). Previously eager execution allocated fresh storage. `lifecycle.md:190` still says only "captured result aliases reusable output storage", and the struct doc at `runner.rs:100-103` says the same.

Separately, `lifecycle.md:198` **removed** the disclaimer "The pilot does not promise allocation-free eager execution" in the same PR that made eager execution session-allocated — the text moved toward a stronger promise than the code makes.

Also note `GraphWorkspace::new` uses `CudaBuffer::alloc` (not zeroed, `workspace.rs:30`) while the session-less path uses `alloc_zeros_async` (`workspace.rs:356`), so a partially-written output differs between the two paths. The bench's eager reference runs session-less, so `EAGER_GRAPH_MIN_COSINE` is not exercising the runner's actual eager contract.

### S4. `DType::I8/I32` feature gate mismatch forces edits in an unrelated family

`apxinf-cuda/Cargo.toml:16` now enables `apxinf-core/quantized-dtypes`, which adds `DType::I8 | I32` for every legacy consumer. That forced new exhaustive-match arms in a non-PI0.5 family (`pi0fast/weights.rs:375-380`, `:409-414`), gated on `#[cfg(feature = "cuda")]` — but the variants are gated on `apxinf-core/quantized-dtypes`, a different feature in a different crate. These agree today only because of how the feature graph happens to resolve. The workspace `Cargo.toml` sets no `resolver`, so a `cargo build --workspace` without `--features cuda` pulls in `apxinf-cuda` (a member), enables `quantized-dtypes`, and compiles `apxinf-model` without `cuda` — at which point those matches are non-exhaustive. Please confirm on a machine with nvcc; if it reproduces, propagate a real `quantized-dtypes` feature through `apxinf-model` instead of proxying on `cuda`.

### S5. Several doc edits contradict the shipped tree

In the same PR: `doc/model-lifecycle.md:228` replaces `apxinf-cuda/` with `apxinf-cuda-new/`, `doc/design.md:68,108,119` calls the legacy crate "the former CUDA backend", and `doc/model-execution-wiring.md:60-62` redirects the operator search order to `apxinf-cuda-new/src/ops/` only — while `doc/model-layer-architecture.md:76` (also edited here) calls the same crate the "Legacy global default" for five families. The legacy tree still exists and is still in use. Please make these consistent.

The rule relaxations in `doc/adding-new-kernels.md` (`:296,321,323,325` — all-candidate tests narrowed to "every changed candidate", alignment/workspace narrowed to "contract boundaries", fallback allow/deny narrowed to "Prepared lifecycle only") are defensible given the direct/prepared split, but they are rules rewritten to fit new code. Please call that out explicitly in the PR description so it is reviewed as a deliberate policy change rather than a cleanup.

---

## Follow-up (not blocking this PR, but please file or note)

**F1. 1365 lines of allocation glue in `blocks/*`, 78% duplicated between precisions.** The new out-param API is the right call — outputs must come from the session arena for capture safety. But the wrappers written to absorb it, e.g.

```rust
// blocks/bf16.rs:49
fn rms_bf16(ctx: &Context, input: &Tensor, weight: &Tensor, eps: f32) -> Result<Tensor>
```

reconstruct the old backend signature (`norm::rms_bf16`) verbatim, in the model layer, once per precision. Comparing the helper regions of `bf16.rs:49-530` and `int8_dynamic.rs:35-506` after normalizing the `l3_` prefix and error strings: 376 of 482 lines are identical. `bf16_vector_prefix` exists three times. `ctx.allocate_output` is already a `Context` method, so this wrapper layer belongs in cuda-new or in `pi05/backend.rs`, written once. As written, a fourth precision means a fourth copy. This also contradicts the "Model mathematics has one owner" review check.

**F2. Shadow workspace formula.** `config.rs:193-302` hand-replays the allocation *sequence* of the blocks, pinned by magic constants (`config.rs:616-625`, e.g. `2_954_794_880`). Nine `allocate(...)` lines had to be added in this PR to keep up. Any future block edit that adds an allocation and forgets this file fails at prepare time with "static inference workspace exhausted". Related: `WorkspaceRequirements.fp8_scratch` was deleted and `Pi05Config::fp8_emulation_scratch_elements` (`config.rs:310`) is now reachable only from its own tests — plausibly intentional given `ensure_supported` rejects FP8 below sm_100, but `doc/pi05-cuda-regression.md:194+` still lists Orin FP8 cells.

**F3. Tests that assert source text.** `crates/apxinf-cuda-new/src/ops/tests/architecture_lifecycle.rs:49,63` and `crates/apxinf-model/tests/cuda_backend_routing.rs:23` grep source strings rather than exercise behavior; they duplicate `check_kernel_architecture.sh` and pass on a comment containing the string. Also, every op test calls `CudaContext::new(0).unwrap()` (17 sites in `independent_operator_semantics.rs`) without an `#[ignore]` or skip, so the binary fails outright on a device-less machine.

**F4. `pi05_bench.rs` reference validation was gutted.** Event-based timing with `host_latency_ms` retained is an honest improvement. But the FP8 integrity schema check (`apxinf.pi05.integrity.v1` — num_views / token_count / action_horizon / action_dim / flow_steps, `fixture == "zeros"`) was deleted, so any JSON with `raw_actions` is now accepted, and the "a raw-image fixture cannot be validated against the zero-input reference" guard is gone. `--views` / `--action-horizon` now override a real checkpoint's config, so a 3-view latency number can be produced from a 2-view checkpoint.

**F5. Audit scripts dropped a check rather than migrating it.** `scripts/audit_pi05_libero_campaign.py` and `audit_pi05_libero_integrity.py` delete the `0 < token_count <= 200` validation; prompt-length bounding is now unverified. `tests/test_audit_pi05_libero.py` covers only the happy path for both auditors — no negative case asserts a rejection.

**F6. Plan-key loose ends.** `GemmPolicy.online_tune` / `cache_dir` are selection-relevant but omitted from `ExecutionKey` (`gemm_execution.rs:14-33`). `runner.rs:571` still maps a now-unreachable `Invalidated` status. Neither appears to be reachable as a stale-reuse bug today, but they are worth closing.

**F7. Calibration seam.** `calibration_amax` (`runner.rs:589-597`) builds a fresh full plan (multi-GiB arena) per call, relies on two different traversals emitting an identical GEMM sequence, and the observer does a blocking `to_cpu` per site inside the session (`calibration.rs:174`). Only the ignored CUDA test `rgb_calibration_reuses_prepared_eager_traversal` covers it.

---

## Summary

B1 is a memory-safety hazard with a small fix. B2 is the one genuine architectural defect — an operator without a fallback pushing its admission rule into the model layer — and is the item I'd most like resolved inside this PR. B3 is the reason two of the three "fix" commits here exist at all. B4 and B5 are policy decisions that need to be made explicitly rather than absorbed into a migration PR.

Also: please consider splitting future work of this size. The delete-and-restore of ~1.3M lines in the branch history made per-commit review impossible, which is part of why the causality regression went unnoticed for 55 commits.

Happy to pair on B1 and B2 if useful.

ChenxiHe231:
## PI0.5 CUDA-new final-HEAD validation (`e35795a`)

All results below were produced from PR HEAD `e35795ad524179e13a8857d42cf429216dcab423`, using the real PI0.5 checkpoint, fixed raw fixtures/reference actions, 10 warmups, and 30 measured graph replays per cell.

### NVIDIA Thor — BF16 and FP8

Artifact: `/home/wwxq/experiments/apxinf-thor-final-e35795a-20260924`

| Precision | Scenario | P50 (ms) | P95 (ms) | Reference cosine | Relative L2 | Eager = graph | Result |
|---|---|---:|---:|---:|---:|:---:|:---:|
| BF16 | 2-view / T10 | 70.637 | 70.772 | 0.9999928572 | 0.003921221 | bitwise | PASS |
| BF16 | 2-view / T21 | 69.867 | 70.097 | 0.9999922156 | 0.004110214 | bitwise | PASS |
| BF16 | 3-view / T10 | 85.927 | 86.254 | 0.9999907076 | 0.004471224 | bitwise | PASS |
| BF16 | 3-view / T21 | 86.574 | 86.717 | 0.9999898745 | 0.004668511 | bitwise | PASS |
| FP8 | 2-view / T10 | 41.701 | 48.852 | 0.9992540007 | 0.039667870 | bitwise | PASS |
| FP8 | 2-view / T21 | 42.688 | 49.864 | 0.9991917109 | 0.040215935 | bitwise | PASS |
| FP8 | 3-view / T10 | 56.162 | 56.440 | 0.9992035997 | 0.040725774 | bitwise | PASS |
| FP8 | 3-view / T21 | 56.538 | 56.787 | 0.9991248729 | 0.042418538 | bitwise | PASS |

README FP8 P50 values are `41.159 / 42.079 / 53.636 / 55.506 ms`; final-HEAD deltas are `+1.32% / +1.45% / +4.71% / +1.86%` respectively. All 8 cells passed reference accuracy and eager/graph bitwise equality. Observed GPU clocks during the run were 315–1575 MHz.

Thor binary SHA-256: `0cbe736941074d08b943dc7519be5e8ebfc950451a58b831248dbb6fe03200d3`.

### Jetson Orin — BF16

Artifact: `/home/wwxq/formal/results/orin-accuracy-perf-e35795a-20260923`

| Scenario | P50 (ms) | Reference cosine | Relative L2 | Eager = graph | Result |
|---|---:|---:|---:|:---:|:---:|
| 2-view / T10 | 162.655 | 0.9999869177 | 0.005115516 | bitwise | PASS |
| 2-view / T21 | 163.850 | 0.9999849961 | 0.005562462 | bitwise | PASS |
| 3-view / T10 | 199.444 | 0.9999837806 | 0.005768445 | bitwise | PASS |
| 3-view / T21 | 199.427 | 0.9999765256 | 0.006971386 | bitwise | PASS |

All 4 cells passed reference accuracy; eager/graph `max_abs = 0`. Model serial regression: `129 passed / 0 failed / 4 ignored`.

Benchmark SHA-256: `4f419bc264d5c22ed442c75e8b26c3e0ec9c64005706c92ad32c4de272350954`.

### RTX 4090 — BF16 paired/interleaved comparison

Artifact: `/tmp/pi05-4090-b98b0cc-vs-e35795a-final-dynamic`

The old-runner baseline's `host_latency_ms` is compared with final HEAD's `latency_ms`; both represent the same host-wall graph-replay boundary.

| Scenario | `b98b0cc` P50 (ms) | `e35795a` P50 (ms) | Delta | Reference cosine | Relative L2 | Eager = graph | Result |
|---|---:|---:|---:|---:|---:|:---:|:---:|
| 2-view / T10 | 33.343 | 31.707 | -4.91% | 0.9999922589 | 0.003969704 | bitwise | PASS |
| 2-view / T21 | 33.343 | 31.846 | -4.49% | 0.9999949213 | 0.003229193 | bitwise | PASS |
| 3-view / T10 | 41.916 | 41.636 | -0.67% | 0.9999900911 | 0.004644163 | bitwise | PASS |
| 3-view / T21 | 41.861 | 41.668 | -0.46% | 0.9999825708 | 0.005904495 | bitwise | PASS |

All 4 final-HEAD cells passed reference accuracy and eager/graph bitwise equality. FP8 is unsupported and rejected on RTX 4090 as intended.

This host did not permit application-clock control. The run is therefore a paired/interleaved dynamic-clock comparison, not a strict frequency-locked measurement; telemetry recorded graphics clocks of 210–2760 MHz and memory clocks of 405–10501 MHz.

Final-HEAD binary SHA-256: `b3749d87171f2dfcc135768372457a44bb6f0c36b775bebde463e4ecce73dce0`.

### Conclusion

- Thor: 8/8 BF16+FP8 performance/accuracy cells passed; eager/graph outputs are bitwise equal.
- Orin: 4/4 BF16 performance/accuracy cells passed; eager/graph outputs are bitwise equal.
- RTX 4090: 4/4 BF16 performance/accuracy cells passed; final HEAD is faster than the paired baseline in all four scenarios.
- These are all direct final-HEAD (`e35795a`) results. The separate LIBERO-10 500-episode evaluation was intentionally not part of this validation run.

PR #90, opened 2026-09-21
feat(core): define portable backend composition contracts

zealoct:
PI0.5 currently depends on CUDA-specific composition, while Qwen3VL still uses host roundtrips for several tensor layout operations. Define the missing portable operator contracts before migrating either model or adding a native HIP backend.

Add object-safe cast, axis slicing/concatenation, permutation, broadcast and stateless attention interfaces with explicit unsupported defaults. Add reusable shape/device/dtype validation, attention mask and intermediate-precision contracts, and document the PI0.5/Qwen3VL operator inventory, fusion boundary, numerical semantics and canonical-versus-packed weight ownership. Existing model execution paths are unchanged.

This is stage 1 only: new device implementations and portable model wiring belong to stage 2. The PR is based on the independent CUDA KV-reset prerequisite branch, so its diff contains only backend contracts.

Validation: 159 CPU tests passed locally; 30 core unit tests and 6 contract tests passed on Thor ARM64. Baseline (6fe4796) and candidate (03642ca) both built with CUDA 13.0.48 on Thor SM110. Two uncontended fixed-protocol CUDA regression pairs passed, first baseline→candidate then candidate→baseline: each pair had 38 tensor dumps exactly equal, identical greedy sequences, and all 30 timing metrics within the predeclared baseline×1.10+0.05ms gate (atol/rtol 1e-5, 3 warmups, 10 samples). Source, harness, fixture, checkpoint, tactics and environment identities were audited; each side used unchanged binaries across repeats.

PI0.5 host-roundtrip medians were 72.185→79.181ms in the first pair and 74.135→73.938ms in the reverse-order pair. The near-limit first result motivated the repeat; both batches are retained. This establishes the regression gate, not a performance improvement or proof of zero overhead. The original baseline's repeated-request Qwen CUDA 700 is covered by the independent KV-reset prerequisite (#89), included identically on both sides; its graph regression fails before the fix and passes after it.

PI0.5 production/eager/graph actions agree exactly. Repository Qwen HF snapshots have existing numerical differences and one text token difference after EOS; their full upstream provenance is unverified, so full HF parity is not claimed. Candidate versus frozen CUDA baseline passes the complete 10-token comparison without relaxed thresholds.

Reproduction commands, manifests, all raw samples, rejected/contended batches and SHA256 seals are retained in `devlocal/backend-contracts/baseline/protocol.md`, `results/` and `reports/acceptance.md` in the phase-1 worktree. These intermediate artifacts remain Git-ignored per repository policy. Maintained contracts and tests are in this PR. Ready for review; left as a draft pending review of the prerequisite and this staged API design.


zealoct:
Pushed `4b49cfd` addressing review items #3, #4, #5. No execution path changes — all six portable ops still return `UnsupportedOp` until phase 2. Workspace compiles (CUDA backend included); 30 core unit tests + 12 contract tests pass.

**#3 — validators wired into dispatch.** `Backend` now declares only `*_impl` hooks. Callers go through a new `PortableOps` trait, blanket-implemented for every `Backend` including `dyn Backend` (`impl<B: Backend + ?Sized>`). Each entry validates args against `contracts`, dispatches, then re-checks the returned shape. A backend can't override or skip validation, so a contract violation surfaces as a structured error rather than being masked by a missing impl. Renaming to `*_impl` is free now — nothing implements these six ops yet.

**#4 — one unimplemented-op signal.** The seven legacy defaults moved from `Error::Other("<op>: not supported…")` to `UnsupportedOp("<op>")`, readable via the new `Error::unsupported_op()`. Nothing in the workspace matched those message strings.

**#5 — typed validation errors.** `contracts` no longer uses `Error::Other`: shape disagreements → `ShapeMismatch` (with both shapes), dtype rejections → new `UnsupportedDType`, invariant violations with no shape to show → new `Contract(&'static str)`. Validation no longer allocates on the error path.

Also split `float_tensor` into `tensor_storage` (device + extent, any dtype) and `float_tensor` (adds the float restriction), so bit-preserving layout ops accept FP8/INT8 while `cast`/attention reject them.

New tests: validation fires before dispatch; layout ops admit non-float dtypes while `cast` refuses them; each legacy default reports its own op name.

Left for the follow-up protocol series (agreed): #2 (`concat_axis`/`attention` superseding `concat_2d`/`vision_sdpa`) and a `ValidatedMask` newtype to make mask value-validation status type-visible.

